Issue #10, opened 2016-03-05
New Custom Control Action does only find controls that either implement Control or directly derive from AbstractControl


erlend-sh:
_From @ghost on March 24, 2014 19:35_

Reference on googlecode: https://code.google.com/p/jmonkeyengine/issues/detail?id=623

_Copied from original issue: jMonkeyEngine/jmonkeyengine#105_

grizeldi:
Was this still not fixed?

MeFisto94:
Nope,
8bec656 is the commit that added the functionality itself but doesn't seem to be looping through the class hierarchy

grizeldi:
Ok, I'll take a look when I have some time. Shouldn't be too hard to fix.

MeFisto94:
Well actually the code is already provided there. If you have the time feel free to do it, only ensure that you test it before and after implementation :P

grizeldi:
Found the part where I have to add new code. Now I just have to think of the best way to scan the class hierarchy.

MeFisto94:
Take a look at the linked googlecode,
there already **is** the code for that, you basically would only need copy+paste :)

grizeldi:
Yes I see. But I'm learning Netbeans API on the go, so it will take some time to figure out which code to copy pasted :)

grizeldi:
Managed to mitigate the problem by checking if the superclass is on the list of control classes. It works if the subclass is in another package but not if it's in the same package...

Can someone point me how to check if class implements an interface? I have found on the internet that it is possible, but you need an annotation processor... Any ways to do it without one?

Dokthar:
@grizeldi An interface is a fully abstract class (without any code) so "instance of" also works for interfaces. I think you can consider interface as a normal class.

grizeldi:
The problem is that I don't get the actual object but an ElementType...
Never worked with those before...

Dokthar:
maybe you can try to encapsulate the code that look for interfaces, and call that code for each super class.

PR #32, opened 2016-03-14
Fixed the scene editor control problem

grizeldi:
Mitigated the problem described in issue #10. The solution works for classes outside the superclass' package.
It doesn't fix the whole issue but until I find a better solution it should be better than nothing.


grizeldi:
FIXED IT!!! :grinning: 
It took me a while to figure it out but now it displays all children that implement/extend Control/AbstractControl.

MeFisto94:
I'll look into it tomorrow after my exam.
Thank you for your work so far :) 

The old PR had the issue that you could have edge cases taking multiple iterations when the order of the classes was wrong.

grizeldi:
Hehe learning for an exam here too ;)

MeFisto94:
Well this still has the same problem:
When you have the following order:
`A extends B; B extends Control` however the list looks like this: `A, B`. Then it will process A but the list doesn't have B yet. Then it goes to B where it notices that B has a Control..

However I'll implement something based on it :)

grizeldi:
Darn, didn't think of that... I'll try to fix it if you won't be faster.

grizeldi:
But I can't think of a way without recursion... :(

MeFisto94:
Note that every recursion can be turned into a loop (and this is basically what my fix does).
Also Note that you didn't take care of the interfaces in your case.

But thanks for the effort anyway and good luck with your exam :)

PR #50, opened 2016-04-08
Moved strings from codechecks to bundle.properties

grizeldi:
Moved strings from codechecks to a separate bundle as a part of preparation for a larger refactor of codechecks package.


MeFisto94:
Looks good to me, I assume you somehow checked that everything works (i.e. you don't have a typo in the key preventing it from loading?)

grizeldi:
I tested updateHint and tempVarsHint. ReadOnlyPrimitives is bugged right now so can't test it. And I don't know on what should internalHint trigger.

MeFisto94:
Okay, we will see in the future then.
Thanks for your Contribution! :)

PR #86, opened 2016-08-06
Fix nifty schema location

grizeldi:
This enables auto completion in the xml editor again.


MeFisto94:
Hey,
Your Pull Request is fine, I'll merge it when I'm less busy.
The reason for keeping it open is that I want to fix that travis bug your PR exposes first and for that I need time and a really really good mood :smile:

grizeldi:
Ok ;)
Take your time. I already manually fixed the only file I am using from that template.

MeFisto94:
Okay, Travis doesn't update the travis file and simply takes it as it was as the pr was requested so I will create another pull request just to test things out.

Merging this

PR #93, opened 2016-09-06
Updated Bloom filter wizard's name

grizeldi:
Also the Bloom Filter Wizard does not support `SceneAndObjects` GlowMode. I was looking into how to fix it and figured out that it would require adding another JLabel into the menu and another `else if` into the `doCreateFilter()` function of `NewBloomFilterAction`. If no one fixes that I can maybe do it later.


MeFisto94:
Are there more Filters without a proper Name? If so you could add it to the pr, else I'll merge it

grizeldi:
No this one is the only one that has a configuration wizard (configuring GlowMode). All other actions just add a filter directly without any configuration. Although it would be great to have some config for the water filter, because of the sheer amount of options.

MeFisto94:
Feel free to add them then but on the other hand you might not know those parameters upon creation and have to try them out?

grizeldi:
Yeah probably it wouldn't make sense for most parameters because they are for fine-tuning, but some would be good.

MeFisto94:
You can just give it a try.
If it turns out to be not that useful, it doesn't do harm anyway.

grizeldi:
Okay, I might try when I have some time to spare.

PR #97, opened 2016-09-21
Fix source version

### LINKED ISSUE #94: error: diamond operator is not supported in -source 1.6
When creating a new JME project (or JmeTests) I can see that nbproject/project.properties contains:
javac.source=1.6
javac.target=1.6

image

This default setting gives errors like:

warning: [options] bootstrap class path not set in conjunction with -source 1.6
/home/j/3D-briller/jme_proj_3.1-beta1-b002-

SNAPSHOT/JmeTests/src/jme3test/app/TestCloneSpatial.java:176: error: diamond operator is not supported in -source 1.6

So, please change all the new project templates's values to
javac.source=1.8
javac.target=1.8

grizeldi:
Fixes #94, swithced to java 7 since java 8 isnt supported on android.

### DIFF
--- a/BasicGameTemplate/nbproject/project.properties
+++ b/BasicGameTemplate/nbproject/project.properties
@@ -49,8 +49,8 @@ javac.compilerargs=
 javac.deprecation=false
 javac.processorpath=\
     ${javac.classpath}
-javac.source=1.6
-javac.target=1.6
+javac.source=1.7
+javac.target=1.7
 javac.test.classpath=\
     ${javac.classpath}:\
     ${build.classes.dir}


nordfalk:
Thanks!
Now that you are changing the files I think you could just as well select Java 8 as default, as Java 8 is coming to Android before you know... https://developer.android.com/guide/platform/j8-jack.html .

If its a problem for a few android developers they can just go to the settings and switch to Java 7 after creating new projects

grizeldi:
Let's wait and see what @MeFisto94 has to say in this regard.

MeFisto94:
We could also apply this to the tests project.
We could set the value to Java 8 since the SDK comes bundled with the JDK 8, however the question is how the Android Developers see this.

If they see something like "Android does not support Java 8, please set to Java 7 in the Project Properties", then it's no big deal to switch to Java 8.

On the Other hand I believe we could stick with Java 7 since it's no real trouble for those who want to use new Java 8 Features to appropriately change the settings.

What do you guys think @pspeed42 @shadowislord @Nehon ?

grizeldi:
Or maybe when enabling android deployment automatically switch to java 7?
But that's out of my quick fix scope of bugfixes.

MeFisto94:
I don't think that's good for two reasons:
- It's not that easy to implement I guess, since our Android Deployment is only some ant tasks and before that only the properties panel. It's not impossible to hook into the Swing Panel and execute something there, but it's not good.

- What if Java 8 would be supported in the meantime? We would have to revert anything and currently existing projects should be set to java 8.

grizeldi:
Well, let's face it, most of current android phones will never support java 8. In that link up there is clearly written that some/most new features are only for Android nougat (and later when new versions come out). Most phones currently out there will never get Android 7...

pspeed42:
JME's current minimum baseline is Java 7 for the time being so the default
should be Java 7.

shadowislord:
jME 3.2 (master) baseline is Java 8 for desktop and Java 7 for Android.
jME 3.1 baseline is Java 7 for desktop / Android.

PR #102, opened 2016-11-23
Update to Netbeans 8.2 final

Ali-RS:


MeFisto94:
Due to 3.1-beta2 being released our engine build script fails again :/
Could you compile it without a problem locally?

Ali-RS:
Yes, I locally updated to nb 8.2 and compiled from engine master about a month a go and it works fine.

Issue #119, opened 2017-03-16
JmeTests should specify native bullet, not jBullet


stephengold:
In the 3.1.0-stable-FINAL version of the SDK, new JmeTests projects are created with the jbullet library in the library properties. However, the physics tests don't work with jbullet any more. To run physics tests, it's necessary to add the bullet and bullet-native libraries and remove the jbullet library.

In lieu of fixing jbullet, perhaps the SDK should specify native bullet for these projects.

Recent discussion at the Forum:

https://hub.jmonkeyengine.org/t/none-of-the-physics-demos-will-run/38197
https://hub.jmonkeyengine.org/t/jmonkeyengine-3-1-0-how-do-i-download-the-massive-sdk/38183/12

normen:
I don't think the issue is that they don't work with jbullet anymore, they should definitely work with it. The issue is the missing jbullet.jar which is not on maven (its our adapted version).

stephengold:
But the JmeTests project uses Ant, not Maven, and there's a jbullet library installed with the SDK. So I don't think that's the issue. Here's the output:

ant -f "S:\\Git\\jME 3.1\\JmeTests" -Dnb.internal.action.name=run run
init:
deps-jar:
Created dir: S:\Git\jME 3.1\JmeTests\build
Updating property file: S:\Git\jME 3.1\JmeTests\build\built-jar.properties
Created dir: S:\Git\jME 3.1\JmeTests\build\classes
Created dir: S:\Git\jME 3.1\JmeTests\build\empty
Created dir: S:\Git\jME 3.1\JmeTests\build\generated-sources\ap-source-output
Compiling 258 source files to S:\Git\jME 3.1\JmeTests\build\classes
warning: [options] bootstrap class path not set in conjunction with -source 1.7
S:\Git\jME 3.1\JmeTests\src\jme3test\bullet\PhysicsHoverControl.java:187: error: cannot access MotionState
            getMotionState().applyTransform(spatial);
  class file for com.bulletphysics.linearmath.MotionState not found
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error
1 warning
S:\Git\jME 3.1\JmeTests\nbproject\build-impl.xml:521: The following error occurred while executing this line:
[...]

PR #120, opened 2017-03-20
jbullet lib lacks jbullet.jar: use native bullet in JmeTests template

stephengold:
My proposed fix for Issue #119 (JmeTests should specify native bullet, not jBullet). I haven't checked to see if the Android version of JmeTests has the same issue.


MeFisto94:
The Android Template still uses 3.0 library names as I've seen. Fortunately the SDK can perform an auto-upgrade but I will change that accordingly.

Thanks for the Contribution

stephengold:
Thanks for the merge!

PR #130, opened 2017-06-09
Second try to get Ogre script to pass Travis.

mitm001:
First attempt failed Travis check so I created a temp directory as a placeholder folder so I could delete the script first, then upload the new script rather than overwriting existing script. Probably wont make a difference but may work.


MeFisto94:
Sorry for making you wait for so long:
Travis is failing because of a weird problem we sometimes have since we cannot depend on the engine using maven but have to build in on your own. This however means that we sometimes have weird version names leading to this bug.

To recapitulate: Did you upgrade or downgrade the ogre script?
And I think we have an automated way of downloading the script so it could be that it is overwritten each build, this means we would have to change the download path, but I don't remember if that is only true for blender/the jdk

mitm001:
> version seems downgraded but the script version stays the same?

What I think happened is the script being used in 3.1 is from a fork of the original script and they never changed the version, just the blender version in notes. 

The script I tried to upload is directly from the google archive that the fork used and referenced. 
https://code.google.com/archive/p/blender2ogre/downloads

The google script does work with 2.78 but the fork does not. I was able to determine the difference because of the file sizes. The fork size matched the size currently used in 3.1 sdk.

mitm001:
Leave a comment

PR #136, opened 2017-08-13
Another try at helping with the SDK.

mitm001:
Travis does not work. Attached are changes to the SDK.

1st change downgrades the Ogre scrip that currently the SDK use and is not compatible with Blender.
2nd change is creating a new BaseAppState template.

Nothing is getting past travis so do with it what you want.


MeFisto94:
I just want to comment on that travis thing: It's a bug I fight with for quite some time already.
Because we build the engine from source but are in a travis environment, it sometimes builds as crazy versions like 3.2.0-3.1.0-beta1-SNAPSHOT or such.

For some reason this issue is suddenly coming up and/or it looks like pull requests trigger it since it is seen as a "tag" and hence is parsed. I'll see what I can do when I have the time, if not I'll just put in the files and commit them before our next release

MeFisto94:
So I've looked into it and the problem is simply that for pull requests, the engine is now built as:
```
Full Version: 3.2-master-pr-130-29
POM Version: 3.2.0-master-pr-130-SNAPSHOT
```

But our code detects it as (for some reason).
```
Full Version: 3.1-master-pr-130-1755
POM Version: 3.1.0-master-pr-130-SNAPSHOT
```

And due to all these problems, I made the version static to `3.2.0-SNAPSHOT`, but it seems that is PR incompatible. I could now maybe use a wildcard for 3.2, but I am uncertain whether that would prefer mavenLocal then. Anyway, I'll look into that and then travis should work fine

MeFisto94:
Thanks for your contribution! Future PullRequests should not lead to any travis errors anymore, I hope, at least until 3.3.0 is out :D

PR #140, opened 2017-09-22
Polish Bundles

koder95:
Added Polish Bundle files. Internationalization.


MeFisto94:
Hey :)
I appreciate your effort in translating the SDK to Polish!

However, there are some minor problems regarding the form of it: The files in `branding` are re-written by gradle/ant upon build time, so we might want a better solution (so when we change the splash screen resolution, it is changed in all translations).

I don't know if netbeans has something for this (like just a different folder), but for the modules themselves, the localization can definitely happen in that form (different bundle files).

So we will have to figure out how to make the branding internationalized first

PR #141, opened 2017-11-03
Update download-jdks.sh

failfarm:
Changed the oracle string to work with the new format, changed the jdk version to 8u152, and altered the jdk extraction logic for linux as xar is not needed with 8u152 .

### DIFF
--- a/jdks/download-jdks.sh
+++ b/jdks/download-jdks.sh
@@ -3,8 +3,9 @@
 #Author MeFisto94
 set -e # Quit on Error
 
-jdk_version="8u111"
-jdk_build_version="b14"
+jdk_version="8u152"
+jdk_build_version="b16"
+jdk_hash=aa0333dd3019491ca4f6ddbe78cdb6d0
 platforms=( "linux-x64.tar.gz" "linux-i586.tar.gz" "windows-i586.exe" "windows-x64.exe" "macosx-x64.dmg" )
 
 function install_xar {
@@ -62,7 +63,7 @@ function download_jdk {
     then
         echo "<<< Already existing, SKIPPING."
     else
-        curl -L  -s -o downloads/jdk-$1 http://download.oracle.com/otn-pub/java/jdk/$jdk_version-$jdk_build_version/jdk-$jdk_version-$1 --cookie "gpw_e24=http%3A%2F%2Fwww.oracle.com%2F; oraclelicense=accept-securebackup-cookie" #--progress-bar
+        curl -v -o downloads/jdk-$1 -L -b oraclelicense=accept-securebackup-cookie http://download.oracle.com/otn-pub/java/jdk/$jdk_version-$jdk_build_version/$jdk_hash/jdk-$jdk_version-$1 #--progress-bar
         echo "<<< OK!"
     fi
 }
@@ -79,7 +80,7 @@ function unpack_mac_jdk {
     fi
 
     download_jdk macosx-x64.dmg
-
+	
     mkdir -p MacOS
     cd MacOS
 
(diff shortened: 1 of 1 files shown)


MeFisto94:
Hey,
Don't let travis discourage you, it's our everlasting problem of addressing the correct versions:
The thing is both the engine and it seems xbuf are building their version strings and/or requirements based upon the travis environment, as such, xbuf tried to depend on `org.jmonkeyengine:jme3-core:3.2.0-patch-3-pr-141-SNAPSHOT.`, however we only have 3.2.0-SNAPSHOT (we don't have the PR name like that in it, I think).

This would be stuff for another investigation and a fix, however if it's really xbufs dependencies, I don't think we can easily change them.

For your PR itself I only have one objection which I will address in-line

jdks/download-jdks.sh
```diff
@@ -62,7 +63,7 @@ function download_jdk {
     then
         echo "<<< Already existing, SKIPPING."
     else
-        curl -L  -s -o downloads/jdk-$1 http://download.oracle.com/otn-pub/java/jdk/$jdk_version-$jdk_build_version/jdk-$jdk_version-$1 --cookie "gpw_e24=http%3A%2F%2Fwww.oracle.com%2F; oraclelicense=accept-securebackup-cookie" #--progress-bar
```
MeFisto94: I understand why you used -v over -s for debugging purposes/local things, however on travis things like a progress-bar and excessive log spam is not what we want. It doesn't only make things less visible (if it continuosly outputs a progress bar) but also we've had trouble with travis killing our build if you output too many chars (which is why the output of ant build-installers only outputs "."). Fun Fact: If you don't output anything, the build is canceled as well.

I am not sure if I am being overly strict with that, I'll try it locally and see if it's too verbose, but it's a detail we could also fix afterwards actually

jdks/download-jdks.sh
```diff
@@ -89,21 +90,25 @@ function unpack_mac_jdk {
         xar -xf /Volumes/JDK*/JDK*.pkg
         hdiutil detach /Volumes/JDK*
     else # Linux
-        7z x ../downloads/jdk-macosx-x64.dmg > /dev/null
+        7z x ../downloads/jdk-macosx-x64.dmg
```
MeFisto94: Same here: Output isn't discarded and hence also destroys that ">> Building the Mac OSX JDK" flair, however as pointed out above, I might be overengineering things

jdks/download-jdks.sh
```diff
@@ -89,21 +90,25 @@ function unpack_mac_jdk {
...
+        #install_xar
+        #./xar-1.5.2/src/xar -xf JDK*/JDK*.pkg
+    cd JDK*
+    
+    7z x JDK*.pkg 
+    cd jdk1*.pkg
+    
     fi
 
-    cd jdk1*.pkg
```
MeFisto94: This is a mistake though, because on Mac OS you would still have to cd into that folder, so the `fi` has to be before `cd jdk1*.pkg`

jdks/download-jdks.sh
```diff
@@ -89,21 +90,25 @@ function unpack_mac_jdk {
...
-        7z x ../downloads/jdk-macosx-x64.dmg > /dev/null
+        7z x ../downloads/jdk-macosx-x64.dmg 
         # The following seems dependent of the 7zip version. If not they also changed their MacOSX Installer
-        7z x 4.hfs > /dev/null
-        install_xar
-        ./xar-1.5.2/src/xar -xf JDK*/JDK*.pkg
+        #7z x 4.hfs > /dev/null
+        #install_xar
+        #./xar-1.5.2/src/xar -xf JDK*/JDK*.pkg
+    cd JDK*
```
MeFisto94: In theory this could also be done like `7z x JDK*/JDK*.pkg`, however I guess both ways are fine here. I just wanted to note that for the sake of completeness

MeFisto94:
Approved functionality, it does build the jdk's just fine. I'll merge and then fix the comments myself

Issue #149, opened 2018-01-10
Add glsl syntax highlighting


grizeldi:
As it stands, working with glsl .vert and .frag files inside the sdk is a nightmare. There's no syntax highlighting (let alone code completion) and indents are a mess. Every time you press enter your new line will start with 0 indents. Not having auto completion for brackets and curly braces is annoying too.
A lot of those problems (but not all of them) can be fixed by installing some third party netbeans plugin, but shaders are a very important thing and I think we should be supporting them by default.

Nehon:
There is a glsl plugin for netbens afaik
http://plugins.netbeans.org/plugin/46515/glsl-syntax-highlighter
I used it when I was using the sdk and it works fine.

grizeldi:
Yeah I am using that too. However my point is that this is one of the things we should be supporting out of the box (the other being gradle build), but we don't. The plugin doesn't provide a fix for indents or brace autocompletion issue. It just adds syntax highlighting, which makes work easier, but the other problems are still very annoying.

MeFisto94:
The Problem here is that:

- I don't know certainly how to add/remove modules from the SDK (I've not found a way to remove nifty without just deleting it).
- We can't just distribute that plugin, since it's licensed as GPLv3. We would need to write a similar plugin. I know that's a bummer but can't change it there.

A custom plugin would be nice though, since it can also autocomplete jme-isms (all the uniform names).
So let's take this issue as the starting point:

## Everyone is free to give this a try.
I've looked into GLSL Plugins early (1-2 years ago), there is good documentation about custom syntax highlighting.
Darkmonkey actually already provides color schemes for that, one just needs the lexer part.
I think this should be in a separate repo though, you can also test this with stock netbeans probably.

grizeldi:
No real point in making this separate from sdk IMO since everyone else can use the existing plugin.

grizeldi:
Work started here: https://github.com/grizeldi/sdk/tree/glsl
Note though that this is my first time working with netbeans, so I'm learning as I go.

MeFisto94:
@grizeldi Is there anything missing actually?

grizeldi:
It's been a while, but iirc syntax highlighting should work. Indents and
braces are still a pain though.

MeFisto94:
Okay, then I'll only rename/rephase this issue, to clearly point out that we could enhance the existing module with brace matching, idents and stuff :)

Issue #155, opened 2018-01-22
Canoncializing the Code


MeFisto94:
Hey Guys,
Since I've just lost many lines of exhausting work thanks to my Git UI, I thought why not splitting this easy task up on many people, so everyone is only a bit exhausted.

It's also a great chance to learn how to fork projects and creating pull requests without having to solve complex coding problems.

The fix consists of three stages, but I might do the third stage on my own.
For the git commit message, a layout of `1./2./3. Canoncialization of jme3-modulename (#<Issue>)` is preferred, make sure to not include "fix(es)" in your commit message though.
Ex: `2. Canoncialization of jme3-scenecomposer (#155)`.

If you follow that, everyone can see below which modules have passed which stage of canoncialization. 

1. Canoncialization:
- Removing "junk" (that's things like: "license.txt not found", "To change this, see Tools -> Templates", which might even be in some of the xml files)
- Updating/adding the LICENSE (if you find that this comment is present, just update it so it includes 2018 (e.g. `2003-2018`).
```java
/*
 * Copyright (c) 2018 jMonkeyEngine
 * All rights reserved.
 * 
 * Redistribution and use in source and binary forms, with or without
 * modification, are permitted provided that the following conditions are
 * met:
 * 
 * * Redistributions of source code must retain the above copyright
 *   notice, this list of conditions and the following disclaimer.
 * 
 * * Redistributions in binary form must reproduce the above copyright
 *   notice, this list of conditions and the following disclaimer in the
 *   documentation and/or other materials provided with the distribution.
 * 
 * * Neither the name of 'jMonkeyEngine' nor the names of its contributors
 *   may be used to endorse or promote products derived from this software
 *   without specific prior written permission.
 * 
 * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
 * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
 * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
 * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
 * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
[...]

Issue #143, opened 2017-12-04
[Scene Composer] Save button is useless


grizeldi:
Like really. Pretty much whatever I do in scene composer, especially if I'm editing terrain, save button will not be enabled. In most cases save all button will be enabled, however after closing the tab with scene composer even if you clicked save all, it still asks you if you want to save since it wasn't saved.
It got so bad I don't even bother with save all and just close and then select save. That seems to work properly at least, but it is **really** annoying. And this issue has been there for literally years.

neph1:
I've noticed this, and have also recently tackled the same problem in the Macaq Editor. I might have a go at this.

MeFisto94:
I just discovered something.
I changed (rotated) a light. Then I had the bug as described.

Selecting a Node in SceneExplorer made it possible to save as expected though.
Could it be that in your case (terrain, lights) it's the same?

grizeldi:
Will have to try, but that's as annoying to do as just closing the tab and saving via popup.

MeFisto94:
That's true and that's not how I meant this, but it might help narrowing down the issue (as in "we're setting the light as current object, not the j3o" or "the light is not seen as belonging to the j3o, where a mesh is")

MeFisto94:
Note this is also related to #67 and #68

PR #156, opened 2018-01-23
GLSL Support Round 1: Syntax Highlighting

grizeldi:
As discussed in #149, we would need some glsl support. This PR adds syntax highlighting for basic syntax as well as all built in types, variables and functions as of GLSL 4.4. It also highlights jme specific variables found [on the wiki](https://jmonkeyengine.github.io/wiki/jme3/advanced/jme3_shaders.html#jme3-and-opengl-3-4-compatibility). Comes with icons for frag and vert files and defaults to annotation netbeans default for other types, since that one looked the best to me. There's also templates for frag and vert files. I would also like to add variable highlighting, but the lexer doesn't know what is a variable as of right now. If we intend to support code completion, this would have to be added.
I haven't tested the code too much, but I have dropped a couple of complicated-looking glsl files I found on github into it and it seemed to highlight just fine.
screenshot from 2018-01-23 20-14-20


MeFisto94:
Don't worry about travis, that's because the engine built as `3.2.0-master-pr-156-SNAPSHOT` but we're searching for the version `3.2.0-glsl-pr-156-SNAPSHOT`. All of this "dynanmic versioning" is annoying me quite a bit :D But I'll just check it out manually

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/GlslLanguage.java
```diff
@@ -0,0 +1,33 @@
+/*
```
MeFisto94: I know it's annoying, but a License Header as described in #155 would be nice to have. It's only a 5 minute task

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/lexer/GlslKeywordLibrary.java
```diff
@@ -0,0 +1,450 @@
+package com.jme3.gde.glsl.highlighter.lexer;
```
MeFisto94: Missing License as well

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/lexer/GlslKeywordLibrary.java
```diff
@@ -0,0 +1,450 @@
...
+class GlslKeywordLibrary {
+    public enum KeywordType{
+        KEYWORD, BUILTIN_FUNCTION, BUILTIN_VARIABLE, BASIC_TYPE, UNFINISHED;
+    }
+    private static final List<String> keywords = new ArrayList<>(),
+            builtinFunctions = new ArrayList<>(),
+            builtinVariables = new ArrayList<>(),
+            basicTypes = new ArrayList<>();
+    
+    static {
```
MeFisto94: I'm not sure if the Array notation in combination with Arrays.asList() would be easier/more compact and maybe even faster. But for this PR, I don't care about that. Just something to keep in mind if you have to work on this file for code completion or something.
But yeah, I'm not picky on that one.

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/lexer/GlslKeywordLibrary.java
```diff
@@ -0,0 +1,450 @@
...
+                }else
+                    returnType = KeywordType.UNFINISHED;
+            }
+        }
+        for (String keyword : keywords){
+            if (keyword.startsWith(s) && (returnType == KeywordType.UNFINISHED || returnType == null)){
+                if (keyword.equals(s)){
+                    returnType = KeywordType.KEYWORD;
+                    break;
+                }
```
MeFisto94: Another #155 here:
- Every if/else clause should be wrapped with {}'s because it improves readability (you always see braces) and it prevents errors from growing where someone is adding a line into the if which actually isn't part of the if.
- Run Netbeans Source -> Format on this file (spaces between `){` missing, in this line the else is not on the same line as the `}` etc. Just a bit of consistent style here.

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/lexer/GlslLanguageHierarchy.java
```diff
@@ -0,0 +1,35 @@
+/*
```
MeFisto94: License Header

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/lexer/GlslLanguageHierarchy.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package com.jme3.gde.glsl.highlighter.lexer;
+
+import java.util.Collection;
+import java.util.EnumSet;
+import org.netbeans.spi.lexer.LanguageHierarchy;
+import org.netbeans.spi.lexer.Lexer;
+import org.netbeans.spi.lexer.LexerRestartInfo;
+
+/**
```
MeFisto94: A description what a LanguageHierarchy does would be nice, but that's just a note for the future, not relevant for this pr

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/lexer/GlslLexer.java
```diff
@@ -0,0 +1,203 @@
+/*
```
MeFisto94: License

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/lexer/GlslLexer.java
```diff
@@ -0,0 +1,203 @@
...
+        tokenFactory = info.tokenFactory();
+    }
+
+    @Override
+    public Token<GlslTokenID> nextToken() {
+        int c;
+        c = lexerInput.read();
+        thisLineSoFar += (char)c;
+        if (isDigit(c)){
+            while (true){
```
MeFisto94: Instead of endless loops, maybe add something like `while (c != LexerInput.EOF)`.
E.g. what happens if you enter a number with a sudden EOF before the Semicolon, endless loop?
But I guess then isDigit(next) evaluates to false in L39.

Just a thing to keep in mind though, in case hang ups occur

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/lexer/GlslTokenID.java
```diff
@@ -0,0 +1,44 @@
+/*
```
MeFisto94: LICENSE

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/recognizer/GlslDataObject.java
```diff
@@ -0,0 +1,135 @@
+/*
```
MeFisto94: LICENSE

nbproject/project.properties
```diff
@@ -53,7 +53,8 @@ modules=\
     ${project.com.jme3.gde.nmgen}:\
     ${project.com.jme3.gde.docs}:\
     ${project.org.jme3.netbeans.plaf.darkmonkey}:\
-    ${project.com.jme3.gde.ios}
+    ${project.com.jme3.gde.ios}:\
+    ${project.com.jme3.gde.glsl.highlighter}
```
MeFisto94: Keep this line

nbproject/project.properties
```diff
@@ -68,6 +69,7 @@ project.com.jme3.gde.core=jme3-core
 project.com.jme3.gde.core.baselibs=jme3-core-baselibs
 project.com.jme3.gde.core.libraries=jme3-core-libraries
 project.com.jme3.gde.core.updatecenters=jme3-core-updatecenters
+project.com.jme3.gde.glsl.highlighter=jme3-glsl-highlighter
```
MeFisto94: Keep this line

branding/core/core.jar/org/netbeans/core/startup/Bundle.properties
```diff
@@ -1,6 +1,6 @@
 #Updated by build script
 #Sat, 27 Aug 2011 22:46:29 +0200
-currentVersion=jMonkeyEngine SDK 3.1.0-alpha5
```
MeFisto94: Drop these Changes

branding/modules/org-netbeans-core-windows.jar/org/netbeans/core/windows/view/ui/Bundle.properties
```diff
@@ -1,4 +1,4 @@
 #Updated by build script
 #Sat, 27 Aug 2011 22:46:29 +0200
-CTL_MainWindow_Title=jMonkeyEngine SDK PBRisComing-SNAPSHOT
```
MeFisto94: Drop these changes

mitm001:
I have read this wiki page before but haven't studied shaders yet so I am not able to update it until then or unless someone writes about it.

jme3-glsl-highlighter/src/com/jme3/gde/glsl/highlighter/lexer/GlslLanguageHierarchy.java
grizeldi: Sadly I have no idea...

grizeldi:
Regarding licence, the default says that I can find the template in the project properties, but for some reason I can't find anything licence related there.

mitm001:
tools>templates>license or do you mean here on github?

grizeldi:
No, not on the github. Netbeans has file templates which are supposed to include licence. However I can't find a way to set that licence.

mitm001:
In the SDK or Netbeans, RMB select the project then properties>License Headers

From there you can change the licenses or edit them. The templates are located by tools>templates>license from the menu bar also.

grizeldi:
Yep, that's it. The first method is also what default template says to do, however it seems that there's no licence headers option in my project properties. I'll try the toolbar method or if that fails just add them by hand.

mitm001:
You must not be using NetBeans 8.1?
https://stackoverflow.com/questions/38484105/how-to-set-project-license-header-in-a-netbeans-module

grizeldi:
Added licenses to all .java files and ran autoformat over them. I didn't however revert the changes made by gradle while building since my git knowledge is kinda weak in that department. I probably could revert the whole files, but the file that has to be only partially reverted is out of my knowledge scope.

@MeFisto94 this doesn't suffer from #143, just tested it and it works as supposed. No idea what happened in the screenshot.

PR #185, opened 2018-12-29
Removed deprecated Gradle stuff and upgraded Gradle to 4.10.2

tonihele:


tonihele:
I added rootdir to https://github.com/jMonkeyEngine/sdk/pull/185/files#diff-c197962302397baf3a4cc36463dce5eaR552. It didn't work for me otherwise in all conditions. The workdir I guess was different when I used Gradle through the NB. Than through the konsole.

MeFisto94:
Okay, travis failing is my fault again, I was just too lazy :D 
The question is now for your replacement from << to doLast: Do we need it to be doLast or was it just a lazy syntax back then to build a task. Can we do this as regular task or (this is something I remember in the back of my head): Do we need doLast() because only then the task is run everytime?

tonihele:
I honestly do not know are they really needed. I didn't try to interpret the build that much. But now it should be 1:1 with the old with the doLasts. As far as I know.
With the doLast some parts are run only when the task is specifically run right?

MeFisto94:
> With the doLast some parts are run only when the task is specifically run right?
This and on the other hand Gradle doesn't try to be smart with "run or not" I think, but yeah, lets do it like this for now :)

PR #215, opened 2019-12-25
Update Gradle 4.10.2 -> 5.6.4

tonihele:
Seems to work the same as the old


MeFisto94:
As this is part of #216, I'll merge it there :)

PR #216, opened 2019-12-28
Cleanups

tonihele:
Hmm, I lil bit f*ed up with the earlier Gradle PR thing... So it shows up in here, but I guess if the earlier PR is accepted it'll be fine (or this one and the other its then obsolete). I don't know to repair it now.

So basically I tried to get rid of all deprecated methods. The files I opened I did do some makeover (usually what was suggested by the NB 11.2 IDE). Basically newer an fancier syntax brought by the Java upgrades in the past years. Also named some threads I found and some other stuff I feel that is more correct way.

I also fixed the copyright message in the about box which annoyed me. And set the copyright year to future at the same time.


MeFisto94:
FYI: You can fix these kinds of stuff by rebasing to master `git rebase -i upstream/master` (upstream might be incorrect, but should be the remote branch of _this_ repo, not your fork).

Then an editor opens (I set mine to notepad++, much better than some command line tool), and you can then edit the individual commits on your branch. There you can drop some, squash them (combine them into one PR) etc. Technically you could also add a revert commit and squash them into another one, so those disappear (error each other out).
Then you would need to force push (git push --force).

But this is okay, as I am now just going to merge this code as a whole, but you might need this tooling whenever someone requires changes to your PR (and wants you to squash them).

It'll take a while since quite a few files are effected :)

MeFisto94:
Thanks for the changes, they look good.
Actually thinking about it we should really parse our warnings and get rid of all of those.
I don't know if using .iterator() instead of for(a:b) throws a warning?
Maybe one day I'll set up some code analysis tool which might also open up issues for every old style file and add them to a milestone or something :P 

But for now, we've improved the code a large bit, thanks :)

tonihele:
Iterators are fine. Not a source of warnings. It is just when you are
looping through the whole thing (and not removing items etc), for each
syntax is preferred, by IDE (and me). For eachs and try withs are new to
those parts of code I guess.

I couldnt remove all jme engine related deprecations for example. So there
are still warnings yesssssh.

PR #218, opened 2019-12-29
Updated build stuff

tonihele:
In all honesty... This is some automatic stuff made by maybe NB 11.2. Looks like it upgraded some of the build scripts. I've build the thing with these no problem, these seemed like an improvement so I didn't discard the changes, instead I offer them here.

Updates some copy lib version and I guess some minimum Java version went from 4 to 6? Also some mentions about Java version 9. So... a good thing?


MeFisto94:
This looked good after local testing, thanks!

PR #219, opened 2019-12-29
Added some generics and removed usages of deprecated methods

tonihele:


jme3-templates/src/com/jme3/gde/templates/basic/BasicGamePanelVisual.java
```diff
@@ -110,7 +110,7 @@ private void browseButtonActionPerformed(java.awt.event.ActionEvent evt) {//GEN-
         String command = evt.getActionCommand();
         if ("BROWSE".equals(command)) {
             JFileChooser chooser = new JFileChooser();
-            FileUtil.preventFileChooserSymlinkTraversal(chooser, null);
```
MeFisto94: I am not sure about that, why has it been removed and what could be a replacement?

jme3-tests-template/src/com/jme3/gde/templates/tests/testsandroid/JME3TestsAndroidPanelVisual.java
```diff
@@ -110,7 +110,7 @@ private void browseButtonActionPerformed(java.awt.event.ActionEvent evt) {//GEN-
         String command = evt.getActionCommand();
         if ("BROWSE".equals(command)) {
             JFileChooser chooser = new JFileChooser();
-            FileUtil.preventFileChooserSymlinkTraversal(chooser, null);
```
MeFisto94: Same here. Does setting the current directory to null prevent symlinks (Note: I didn not read the javadoc, I don't know what the methods do)

jme3-tests-template/src/com/jme3/gde/templates/tests/testsdesktop/JmeTestsPanelVisual.java
```diff
@@ -110,7 +110,7 @@ private void browseButtonActionPerformed(java.awt.event.ActionEvent evt) {//GEN-
         String command = evt.getActionCommand();
         if ("BROWSE".equals(command)) {
             JFileChooser chooser = new JFileChooser();
-            FileUtil.preventFileChooserSymlinkTraversal(chooser, null);
+            chooser.setCurrentDirectory(null);
```
MeFisto94: and here

tonihele:
http://bits.netbeans.org/dev/javadoc/org-openide-filesystems/org/openide/filesystems/FileUtil.html

Apparently that method was there to fix a bug that existed in Java 5 and
earlier. The link says to use the other method instead. Im not sure if it
is needed or not, it might be the default behavior anyway. For me it opens
to My documents in Windows.

Issue #37, opened 2016-03-19
Upgrade "Code Checks" to use Hint over AbstractHint


MeFisto94:
The Code Checks Module is relying on a deprecated Netbeans API:

```
jme3-code-check/src/com/jme3/gde/codecheck/hints/TempVarsHint.java:20: warning: [deprecation] AbstractHint in org.netbeans.modules.java.hints.spi has been deprecated
[ant:nb-javac] import org.netbeans.modules.java.hints.spi.AbstractHint;
```

The thing is, we have to upgrade from `extends AbstractHint` to `implements Hint`
See [here](http://www.grepcode.com/file/bits.netbeans.org/maven2/org.netbeans.api/org-netbeans-spi-java-hints/RELEASE80-BETA/org/netbeans/spi/java/hints/Hint.java#Hint)

Actually we don't have to Implement the Hint but rather uses it's Annotations.
This unfortunately is more than a small refactoring. See [here](https://platform.netbeans.org/tutorials/nbm-java-hint.html)

It seems to be Deprecated in 8.0 so it's really low priority, but when you want to learn a bit about Netbeans API and do some chilled refactoring, this one is for you :)

grizeldi:
This one looks interesting. I'll try to do it when I have some time.

grizeldi:
Just checked the code. I'll see what I can do.
And if I'm already working on it, can I move the strings to bundle.properties?

MeFisto94:
Sure you can, this is actually even prefered since it allows you to add localization to the SDK

grizeldi:
When is ReadOnlyPrimitives supposed to fire? Can't get it to work on geometry.getLocalTranslation().set(blah blah blah). Maybe I screwed up something in the part where I moved id strings to the bundle. I'll check tomorrow.

grizeldi:
Hmm... Checked the id string and it should work. Now I guess I'll try to build a clone of this repo to see if it works here. Otherwise I just found a new bug.

theom:
This still seems to be an issue so I'll try and take a stab at this.

theom:
Sorry, bailing out. This is too big of a chunk for me at the moment.

Some pointers to keep in mind when fixing this:

* If you are new to NetBeans hints then I strongly suggest you go through the [tutorial](https://platform.netbeans.org/tutorials/nbm-java-hint.html) MeFist94 mentions. It will at least get you going in the right direction.
[...]

PR #221, opened 2020-01-01
Remove legacy hints module, rewrite the hints to the newer module

tonihele:
Phew, renewed the hints. I didn't do the unit tests. I tried to, but I get some compilation problems with them and the test wont work. This should be looked into in the future. This is anyway kind of major deprecation removal.

https://platform.netbeans.org/tutorials/nbm-java-hint.html

If you just do them by the instructions, a unit test is added. Looks totally worth writing, as the manual testing is rather tedious.

Also the ReadOnly hint was broken I think. It didn't trigger on "set" method as it should. However, both ReadOnly and Internal are not in use in JME currently (Internal is but not on a method). I tested them with custom classes.


tonihele:
Test class

```
package mygame;

import com.jme3.app.SimpleApplication;
import com.jme3.material.Material;
import com.jme3.math.ColorRGBA;
import com.jme3.math.Vector3f;
import com.jme3.renderer.RenderManager;
import com.jme3.scene.Geometry;
import com.jme3.scene.Node;
import com.jme3.scene.shape.Box;
import com.jme3.system.Annotations;
import com.jme3.util.BufferAllocatorFactory;
import com.jme3.util.TempVars;
import javax.swing.JOptionPane;

/**
 * This is the Main Class of your Game. You should only do initialization here.
 * Move your Logic into AppStates or Controls
 * @author normenhansen
 */
public class Main extends SimpleApplication {

    public static void main(String[] args) {
        Main app = new Main();
        app.start();
    }

    @Override
    public void simpleInitApp() {
        Box b = new Box(1, 1, 1);
        Geometry geom = new Geometry("Box", b);

        Material mat = new Material(assetManager, "Common/MatDefs/Misc/Unshaded.j3md");
        mat.setColor("Color", ColorRGBA.Blue);
        geom.setMaterial(mat);

        rootNode.updateModelBound();
        TempVars derp = TempVars.get();

        TempVars derpy = TempVars.get();
                JOptionPane.showMessageDialog(null, null);
               // com.jme3.scene.Node.updateGeometricState();
rootNode.getLocalTranslation().set(Vector3f.ZERO);
        String derpf = BufferAllocatorFactory.PROPERTY_BUFFER_ALLOCATOR_IMPLEMENTATION;

        ReadOnlyNode ghfgh = new ReadOnlyNode();
        ghfgh.set(null);
        ghfgh.setKey(null);

        ghfgh.get();
        rootNode.attachChild(geom);
    }



    @Override
    public void simpleUpdate(float tpf) {
        //TODO: add update code
    }

    @Override
    public void simpleRender(RenderManager rm) {
        //TODO: add render code
    }

    @Annotations.ReadOnly
    protected class ReadOnlyNode extends Node {

        @Annotations.ReadOnly
        public void set(String string) {

        }

        @Annotations.Internal
        public void get() {

        }

    }
}

```

grizeldi:
This fixes #37 if I understand correctly?

tonihele:
It looks to me that it very much does solve the issue. Finally.

jme3-code-check/src/com/jme3/gde/codecheck/hints/ReadOnlyPrimitiveHint.java
```diff
@@ -1,144 +1,91 @@
...
-            if (fail) {
-                return Collections.<ErrorDescription>singletonList(
-                        ErrorDescriptionFactory.createErrorDescription(
-                        getSeverity().toEditorSeverity(),
-                        getDisplayName(),
-                        NO_FIXES,
-                        info.getFileObject(),
-                        (int) info.getTrees().getSourcePositions().getStartPosition(info.getCompilationUnit(), t),
-                        (int) info.getTrees().getSourcePositions().getEndPosition(info.getCompilationUnit(), t)));
+            if ("set".equals(el.getSimpleName().toString())) {
```
tonihele: This is wrong, it should be startsWith. Like the original tried to do, I assume. I'll fix it after I'm back from work.
```suggestion
            if (el.getSimpleName().toString().startsWith("set")) {
```

jme3-code-check/src/com/jme3/gde/codecheck/hints/InternalMethodHint.java
```diff
@@ -1,134 +1,84 @@
 package com.jme3.gde.codecheck.hints;
 
-import com.jme3.system.Annotations.Internal;
```
MeFisto94: I think we should open an issue on the engine to check the source and add a few of these where appropriate, as I think noone knows about them anymore :D

tonihele:
There, good to go!

As I said, does what it used to. Except ReadOnlyPrimitiveHint was also fixed as the old failed to check the "set", it looked for it from the wrong place.

PR #240, opened 2020-02-14
Fixed download-jdks.sh

buahaha:
Better(?) `if` statement and removed unsupported `-r` argument from `ln` command.


MeFisto94:
The `-r` flag stands for `relative` and is required on some `Ubuntu` Distributions it seems.
When Mac OS X doesn't support it, we either have to live with that (we build on Ubuntu for all platforms) or maybe we can fix that by making the paths absolute by prepending `$(pwd)`.

OR we can just try the same command without -r on Mac OS X.

PR #247, opened 2020-02-28
fix broken iOS sound support

revvv:


MeFisto94:
So did you basically rename the files as they were having the wrong name?
Also EFX looks like it is not supported or at least the methods are commented?

revvv:
I renamed the files from the latest Android native package (see link in commit) and then adapted them:
- add correct header files
- comment non-existing functions

The same was done for the previous sound implementation.
You can see that the same methods were commented, too.

MeFisto94:
Sorry for getting back so late.
I wonder if we can enable the Sound Effect LIbraries somehow?
But that's a matter for another pull request then :)

btw: did you have problems with the aspect ratio in the gui node? When I was testing iOS, it was distorted, but I only could test in the emulator.

revvv:
Thanks!

No, aspect ratio was OK, but I already added a Launch Screen when I started the simulator the first time. Without Launch Screen you get a much smaller resolution on a device. For example 640 x 960 instead of 750 x 1334.

Is it OK to submit three more PRs or should I submit one after the other?

MeFisto94:
Feel free to submit them all at once, we can still review them one at a
time :)
The only exception would be when one PR depends on another, because then
we can do them sequentially and the CI will build successfully.

Thanks for your contributions!

PR #248, opened 2020-03-01
fix iOS touch: added scale factor

revvv:
- jmeAppDelegate.m: without the scale factor (which is device specific) you'll hit the wrong location
- jmeAppDelegate.m: the cast (long) is required for the simulator
- logging.properties: added comments how to turn on logging, this can be really helpful

### DIFF
--- a/jme3-ios/ios-data/templates/project/jme-ios/jmeAppDelegate.m
+++ b/jme3-ios/ios-data/templates/project/jme-ios/jmeAppDelegate.m
@@ -276,8 +276,8 @@ getEnv(JavaVM* vm)
     	UITouch *touch = [touches anyObject];
     	CGPoint position = [touch locationInView: nil];
         float scale = _glview.contentScaleFactor;
-        // NOTE: cast to long is only required for x86_64/simulator, otherwise JmeAppHarness receives non-sense values (Avian!?)
-        (*e)->CallVoidMethod(e, self.harness, self.injectTouchBegin, 0, (long) touch.timestamp, position.x * scale, position.y * scale);
+        // NOTE: cast to jlong is required, otherwise JmeAppHarness receives non-sense values
+        (*e)->CallVoidMethod(e, self.harness, self.injectTouchBegin, 0, (jlong) touch.timestamp, position.x * scale, position.y * scale);
         if ((*e)->ExceptionCheck(e)) {
             NSLog(@"Could not invoke iOS Harness injectTouchBegin");
             (*e)->ExceptionDescribe(e);
@@ -293,7 +293,7 @@ getEnv(JavaVM* vm)
     	UITouch *touch = [touches anyObject];
     	CGPoint position = [touch locationInView: nil];
         float scale = _glview.contentScaleFactor;
-        (*e)->CallVoidMethod(e, self.harness, self.injectTouchMove, 0, (long) touch.timestamp, position.x * scale, position.y * scale);
+        (*e)->CallVoidMethod(e, self.harness, self.injectTouchMove, 0, (jlong) touch.timestamp, position.x * scale, position.y * scale);
         if ((*e)->ExceptionCheck(e)) {
             NSLog(@"Could not invoke iOS Harness injectTouchMove");
             (*e)->ExceptionDescribe(e);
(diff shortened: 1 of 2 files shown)


revvv:
I have no idea why the build has failed.

MeFisto94:
The CI failures are my fault: https://travis-ci.org/jMonkeyEngine/sdk/jobs/656986786#L1263
Actually in order to use 11.2 I had to add a command line which updates Netbeans to the latest version (here in the meantime 11.3) and thus the harness override seems faulty.

I could fix that in master, if you have a strong git-fu, you can rebase your PRs on that and force push so that CI runs again, but if not it's okay as well, because I guess none of the PRs touch files affecting the build anyway.

revvv:
Thanks and good to know it's not my fault, but I haven't seen other checkins...

MeFisto94:
The proper cast would probably be `jlong`, also see here:
https://github.com/mapbox/jni.hpp/blob/385dede669c843144b9ad68bbb7af3989d76104e/test/openjdk/jni.h#L42-L43
The size of a long depends on the architecture. I think only on linux 32 bit, long is 32bit, and everywhere else long = long long = 64bits, see:
https://github.com/mapbox/jni.hpp/blob/master/test/openjdk/jni_md.h#L34-L38

Now that way, your code would actually "break" on 32bit architectures, as you only pass the lower half of the timestamp.

Casting might've become necessary with an Apple API Change, see: https://developer.apple.com/documentation/uikit/uitouch/1618144-timestamp

This however is a Double, see: https://developer.apple.com/documentation/foundation/timeinterval
I guess when you pass a double to a java method expecting a long, the method call probably failed.
What was the old behavior? Was an exception thrown? Or could we need more error checking here?

revvv:
Without the cast, there was no exception, just the wrong touch location was submitted, but only in the x86_64 simulator. On a iOS 64 bit device (my iPhone 6) the correct touch location was submitted, but a wrong timestamp, which didn't harm. I don't use it.

I did these tests now:
```
Test case: Touch at down left
 
 On x86_64 simulator
-----------------------
 Test 1: w/o cast
 iOS:  timestamp=152632.369324   x=9.000000  y=1331.000000
 Java: timestamp=140735507257888 x=152632.38 y=9.0
 
 Test 2: w/ cast to long
 iOS:  timestamp=152947.991889 x=9.000000 y=1333.000000
 Java: timestamp=152947        x=9.0      y=1333.0
 
 Test 3: w/ cast to jlong
 iOS:  timestamp=153071.750553 x=7.000000 y=1327.000000
 Java: timestamp=153071        x=7.0      y=1327.0
 
 On iPhone 6
--------------
 Test 1: w/o cast
 iOS:  timestamp=725230.854457       x=34.000000 y=1301.000000
 Java: timestamp=4694476897696575817 x=34.0      y=1301.0
 
 Test 2: w/ cast to long
 iOS:  timestamp=725354.845563 x=50.000000 y=1286.000000
 Java: timestamp=725354        x=50.0      y=1286.0
 
 Test 3: w/ cast to jlong
 iOS:  timestamp=725432.918414 x=37.000000 y=1312.000000
 Java: timestamp=725432        x=37.0      y=1312.0
```

As you see, in the simulator the iOS timestamp messes the Java timestamp and the x value. And y contains the x value.

A cast to jlong would be probably the best solution. Tell me, if you should commit it.

MeFisto94:
Well, I guess we have a bad corner case there, where without the cast, the timestamp is seen as the X parameter. I wonder what the timestamp is there, probably some pointer.

I guess Apple doesn't allow 32bit anymore? Technically speaking `jlong` isn't required there anymore, but please include it for the sake of professionality/completeness.

Otherwise I'd be curious to see the output of a 32bit simulator, because there, timestamp probably is faulty when using a `long` cast.

And thanks for taking the time! :)

revvv:
You're welcome. I committed the cast to jlong.

Indeed with latest Xcode you cannot submit 32 bit apps. When I submitted my 32 bit apps some years ago, there was no difference between device or simulator, but I haven't checked the timestamp.

MeFisto94:
I finally found some link: https://software.intel.com/en-us/articles/size-of-long-integer-type-on-different-architecture-and-os
Unfortunately it does not talk about ARM, but it's funny to see how long seems to be a completely random size :D

Thanks again!

revvv:
Thanks! And thanks for providing all the links. Now I know better...

PR #249, opened 2020-03-01
iOS: add ProGuard exclusions

revvv:
1. Exclusions:
java.nio.charset.Charset:
Otherwise you are limited to UTF-8 (and for example cannot use java.util.zip, because it uses ISO-8859-1)

For java.util.zip.Deflater:
java.util.zip.Deflater
java.io.FileNotFoundException

For reflection (IIRC nifty-gui and Java Annotations)
java.lang.reflect.Proxy
java.lang.reflect.InvocationHandler

For logging:
java.util.logging.ConsoleHandler
java.util.logging.FileHandler
java.util.logging.SimpleFormatter
java.util.logging.LogManager

For nifty-gui:
org.xmlpull.mxp1.MXParserFactory

2. Removed iOS version: just use link to the most recent version
This should reduce failing builds with new projects significantly!


MeFisto94:
Thank you very much! :)
Just one hint: 1 and 2 could've been separate pull requests as they are about separate things

revvv:
Thanks. Ok, next time even more PRs... ;-)

PR #254, opened 2020-03-25
Add multi-touch support for iOS

revvv:
The current implementation only forwards one touch. Now I iterate through the list.
I also use a pointer ID instead of 0, so that it's possible to distinct between the fingers.
The cast to (jint) is optional I think.

The length of the list indicates how many fingers started the touch, so now we can detect gestures. (Which requires custom code, of course.)

### DIFF
--- a/jme3-ios/ios-data/templates/project/jme-ios/jmeAppDelegate.m
+++ b/jme3-ios/ios-data/templates/project/jme-ios/jmeAppDelegate.m
@@ -273,15 +273,18 @@ getEnv(JavaVM* vm)
     NSLog(@"touchesBegan");
     JNIEnv* e = getEnv(self.vm);
     if (e) {
-    	UITouch *touch = [touches anyObject];
-    	CGPoint position = [touch locationInView: nil];
-        float scale = _glview.contentScaleFactor;
-        // NOTE: cast to jlong is required, otherwise JmeAppHarness receives non-sense values
-        (*e)->CallVoidMethod(e, self.harness, self.injectTouchBegin, 0, (jlong) touch.timestamp, position.x * scale, position.y * scale);
-        if ((*e)->ExceptionCheck(e)) {
-            NSLog(@"Could not invoke iOS Harness injectTouchBegin");
-            (*e)->ExceptionDescribe(e);
-            (*e)->ExceptionClear(e);
+        NSEnumerator *enumerator = [touches objectEnumerator];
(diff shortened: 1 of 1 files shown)


jme3-ios/ios-data/templates/project/jme-ios/jmeAppDelegate.m
```diff
@@ -273,15 +273,19 @@ - (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event {
...
-            NSLog(@"Could not invoke iOS Harness injectTouchBegin");
-            (*e)->ExceptionDescribe(e);
-            (*e)->ExceptionClear(e);
+        NSEnumerator *enumerator = [touches objectEnumerator];
+        UITouch *touch;
+        while (touch = [enumerator nextObject]) {
+            CGPoint position = [touch locationInView: nil];
+            float scale = _glview.contentScaleFactor;
+            //NSLog(@"multipleTouchEnabled=%d", _window.rootViewController.view.multipleTouchEnabled); // true
+            // NOTE: cast to jlong is required, otherwise JmeAppHarness receives non-sense values
```
MeFisto94, line 282: It is unclear what this logging statement should be/why it is commented.
The second comment should probably be copied to every related CallVoidMethod statement

MeFisto94:
Sorry for the very late response!
Thanks for keeping pushing our iOS Code forward :)
When the two small comment-things fixed, we can merge this

revvv:
The debug statement is a left-over: Apple's tutorial says you have to set multipleTouchEnabled to true, but that was already the case. Removed.

Copied cast comment to all relevant blocks.

revvv:
Sorry, I pressed "Resolve conversation." To your code comments. Deleting it was not my intention...

revvv:
Thank you very much for merging in my updates. And take your time. 7 days are not really very long.

MeFisto94:
> The debug statement is a left-over: Apple's tutorial says you have to set multipleTouchEnabled to true, but that was already the case. Removed.

Well, might be good to know or see, but generally a commented out code needs a comment to explain, but it's also good that way.

Do we set the multitouch in the harness or is it something which could go wrong for the user?

revvv:
> Do we set the multitouch in the harness or is it something which could go wrong for the user?

I guess `multipleTouchEnabled = true` is the default, it's not explicitly set.

PR #256, opened 2020-03-29
Remove the Update Center Subproject

### LINKED ISSUE #251: Remove the jMonkeyEngine update repositories
Remove the jMonkeyEngine update repositories. As briefly discussed with @MeFisto94, these were never really used. They should be removed since they cause some error messages when checking for updates. Probably other slowdowns also when connection to these are tried.

tonihele:
Resolves #251 

Complete removal of the project. Doesn't make sense to keep it. If the need arises in the future. Then we come back to this.

After this the Check for updates feature doesn't report an error anymore. Instead a success. Thus gives user the feeling that the SDK is less buggy :)


jme3-core/nbproject/project.xml
```diff
@@ -24,17 +24,11 @@
                         <specification-version>3.3.0</specification-version>
                     </run-dependency>
                 </dependency>
-                <dependency>
-                    <code-name-base>com.jme3.gde.core.updatecenters</code-name-base>
-                    <run-dependency>
-                        <specification-version>3.3.0</specification-version>
-                    </run-dependency>
-                </dependency>
```
MeFisto94, line 32: I am worried by this: Why and how does jme3-core depend on the update centers?

MeFisto94:
And I don't know if we should just delete the folder or exclude it from the build (i.e. this one change in the xml file where it has been removed should be enough).

But on the other hand it makes the code cleaner/easier and in theory git history shouldn't break

jme3-core/nbproject/project.xml
tonihele: Maybe just bundle it to the project? How could this be verified? There is no direct dependency at least. And there shouldn't either.

tonihele:
I first also thought that just exclude. But the facts I think are that it is not coming back anytime soon. And when and if it comes back, the implementation might be even better to start from scratch with fresh instructions and fresh code. It is not even that big amount of code to begin with.

If not deleted then in theory it should be maintained. Even not build. Adds to confusion. So I opted to delete it completely. You are the boss however, so what ever you want...? :)

jme3-core/nbproject/project.xml
MeFisto94: Probably only by testing everything and seeing if one is getting crashes.
Additionally maybe searching through the relevant code.

Specifically I think about some code in the Installer.java files, which are executed on startup, which most certainly add the update centers.

MeFisto94:
> If not deleted then in theory it should be maintained. Even not build. Adds to confusion. So I opted to delete it completely. You are the boss however, so what ever you want...? :)

I don't know, the problem is that the value of the update center tightly depends on the code frequency. It's a pipe dream to believe the SDK will see weekly or even daily updates for the users to test and also the infrastructure was seen critically (by non SDK users!), but in theory we could use that at some point.

But I generally don't feel confident deleting code when we could just disable it (exclude the project from the suite). But then we have git, we can always revert the delete commit in the future (provided we find it again :smile: )

MeFisto94:
I've decided:
We'll disable the module for 3.3 and once we've cut a 3.3 branch, we can start removing several modules from the master branch.

MeFisto94:
I've resolved the conflicts to genfiles.properties.
Chances are that they show up as modified once commited again, but this is non critical

MeFisto94:
Officially rejecting this based on changed plans: We will have a github-releases based update center based completely on Github Actions.

See https://github.com/jMonkeyEngine/sdk/blob/489487f6d4970efebd8a5825cdb668bbf383e9ce/.github/workflows/gradle.yml

which deploys to 

https://github.com/jMonkeyEngine/sdk-update-center/blob/52ebc5f7c56ff2f24ee2eb07f611f075cafd288e/.github/workflows/nightly.yml

which creates
https://github.com/jMonkeyEngine/sdk-update-center/releases/tag/nightly_10
and 
https://github.com/jMonkeyEngine/sdk-update-center/commit/db2cee205cd309128f348a8eed0388a992c84924

PR #257, opened 2020-03-29
Cleanup dependencies

### LINKED ISSUE #237: Warning: GraalVM Debugging Support - Missing truffleapi
Warning - could not install some modules: GraalVM Debugging Support - The module named org.netbeans.libs.truffleapi was needed and not found.

SDK v3.3.0-beta1-sdk1
macOS Catalina 10.15.3

tonihele:
Cleaned up all that I could. This is done against the Netbeans 11.3.

Also I excluded the Truffles. They don't seem to be needed by us. It removes the annoying GraalVM module warning from the startup. Now the experience should be smooth as butter :P

Resolves #237


tonihele:
I originally kinda did this for the Maven conversion.. But that is on suspend now. This will anyway help any such attempts in the future too. Getting rid of this excess garbage..

jme3-android/nbproject/project.xml
```diff
@@ -21,49 +21,15 @@
...
-                        <implementation-version/>
-                    </run-dependency>
-                </dependency>
-                <dependency>
-                    <code-name-base>org.netbeans.modules.java.project</code-name-base>
-                    <build-prerequisite/>
-                    <compile-dependency/>
-                    <run-dependency>
-                        <release-version>1</release-version>
-                        <specification-version>1.62.1</specification-version>
```
MeFisto94, line 66: So this is a big PR, so it'll take me some time, but have you ensured that everything still works?
Specifically this lines already make me wonder: We've used the Java Project Stuff to add our custom ANT Tasks used for Android, iirc.

So I wonder if this isn't breaking anything but only on runtime (run dependencies, no compile dependencies)

tonihele: Good points. I only tested some very basic parts. I can test some more and check in here then.

How the android part can be tested? I've never used it.

MeFisto94: Probably only by going into Android Support in the Project Settings and using the broken SDK Version Selector. Also #260 though

neph1:
I've merged master into this and tested it a bit. Loads of warnings (seems related to java 11 switch), but it builds.

- Scene creation OK
- Node and primitive creation OK
- Material creation and editing OK
- Filter creation and loading OK
- Gltf conversion OK

More to test?

tonihele:
I'll close this in favor of https://github.com/jMonkeyEngine/sdk/issues/367. This is pretty much stale this moment and basically the Maven conversion I would do so that is actually includes this cleaning task by itself. This way we don't risk breaking 3.5 release either.

Issue #255, opened 2020-03-27
Include Gradle plugin installed as default and add the jME Gradle project template


tonihele:
Include Gradle plugin installed as default and add the jME Gradle project template (I think the store or the jME documentation had this?). We would still have the ye olden Ant template. But also offer some modern bells & whistles. This would also a bit decouple the SDK from the engine. As the template allows even inexperienced user to use jME version they want.

swiftech:
The Gradle or the Maven template is better than the  Ant template to me

MeFisto94:
Using the Gradle Template would remove all the SDK Functionality from the User, to the point of being able to use the Gradle Template with another IDE.

So we first need to fix Gradle on the SDK (see the branch gradle-support, the commit message is pretty explainatory)

swiftech:
I have tried the latest preview build for Gradle template, but I didn't find any Gradle things after creating a project by the BasicGame template. Did I miss something to enable this functionality?

tonihele:
This issue is still open (not implemented). The preview should have Gradle plugin enabled by default. So you don't have to manually install it. The BasicGame template is still Ant template as far as I know. So either open an existing Gradle project or create new one, but no template available.

MeFisto94:
Exactly, you need to open a "jme gradle project", that is a project with gradle which contains a subproject called `assets`.

There everything else should work: Asset Editing, Starting, Debugging.
The template will be more complicated as I am thinking about including Android and have it Android Studio compatible.

tonihele:
This is implemented some time ago.

PR #259, opened 2020-04-04
Enable Gradle support by default

tonihele:
Adds Gradle support by default. Also adds support for Maven, I guess that is alright. This is like the minimum amount of stuff brought in with the Gradle.

Next up could be implementing the #255 template. But at least this is a nice addition on its own.

### DIFF
--- a/nbproject/platform.properties
+++ b/nbproject/platform.properties
@@ -3,6 +3,7 @@ keystore=../nbproject/private/keystore
 nbm_alias=jmeupdates
 cluster.path=\
     ${nbplatform.active.dir}/extide:\
+    ${nbplatform.active.dir}/groovy:\
     ${nbplatform.active.dir}/harness:\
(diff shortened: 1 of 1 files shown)


MeFisto94:
And you can launch the build SDK like that?!
Because I think that was where I was already failing, but actually being more specific:
When adding a module which depended on those gradle things, it refused loading.

See https://github.com/jMonkeyEngine/sdk/commit/ab1cd3b011aa65a8fdc0f0fcad3019daa85f439f
BUT: I forgot to add the gradle plugins in that config for some reason.

I'll give this a try

tonihele:
I launch from plain NB 11.3 the SDK project. There the old ant project template, test suite and OpenKeeper (a supreme Gradle based project) works out of the box.

MeFisto94:
Closing as I've added that into the Gradle-Support now, I guess it helped me, so thanks!

PR #263, opened 2020-04-05
Cleanups 3

tonihele:


MeFisto94:
Did the IDE auto-suggest this property changes or was:
> // TODO if using org.openide.dialogs >= 7.8, can use WizardDescriptor.PROP_*:

Your "inspiration"? Because if you manually did this, we should at least fire every Wizard up once and see if it crashes or works (because I just had to deal with properties being unable to be loaded, yesterday)

tonihele:
It was in fact that TODO item there. I believe this is just compile time magic. Magic strings changed to constants provided in the modules. And the constants do have the same string. I can do some further tests this week.

MeFisto94:
According to https://github.com/apache/netbeans/blob/master/platform/openide.dialogs/src/org/openide/WizardDescriptor.java you are right

PR #269, opened 2020-04-11
iOS: enable nifty-gui support

revvv:
Without the ProGuard exceptions at least these elements are not usable:
ButtonBuilder, LabelBuilder, TextFieldBuilder

The exceptions were hard to find: iOS log does not show what's missing.
I only saw that nifty-gui doesn't find the control files.

But then I remembered I have previously added some exclusions for Android, and just copied them to iOS and vóilà it works. (logcat showed which classes were missing.)


MeFisto94:
As always thanks for the PR! :)

revvv:
You are welcome! Thanks for merging!

PR #272, opened 2020-04-19
Fixed some documentation links

tonihele:


jme3-core/src/com/jme3/gde/core/scene/SceneApplication.java
```diff
@@ -715,12 +712,12 @@ public static void showStartupErrorMessage(Throwable exception) {
         NotifyUtil.show("Error starting OpenGL context!", "Click here to go to troubleshooting web page.", MessageType.EXCEPTION, lst, 0);
         logger.log(Level.INFO, exception.getMessage(), exception);
```
MeFisto94, line 713: Actually this could also be changed into ExceptionUtils.show(??) for the user to have a nice exception msg

MeFisto94:
So I am going to merge it, no matter the comment, actually using ExceptionUtils is worse, because then the user can't directly open a browser, which this code seems to do.

PR #274, opened 2020-06-01
Upgrade optlibs: Heart, Minie, and Wes

stephengold:


MeFisto94:
Thanks for helping to maintain the libs!

PR #281, opened 2020-08-08
Enable to display CJK characters for sdk v3.2

### LINKED ISSUE #73: 3.1-beta1-SNAPSHOT can't display chinese charset
http://chuantu.biz/t5/7/1465529422x1035372876.png

hangingman:
fix #73

## Whats' this PR do ?
- Fix #73 issue, enable to display CJK characters for sdk v3.2

image

## Problem

- According to this StackOverflow post, DejaVu font doen't support CJK glyphs
  - https://stackoverflow.com/questions/18065028/does-dejavu-has-support-for-cjk-chinese-japanese-korean-glyphs/18065154#18065154
- However, jMonkey sdk, `DarkMonkeyLookAndFeel.java` forces swing window to use font as "DejaVu Sans Condensed-PLAIN-12".
- This causes menu displays to be unreadable for CJK locale users ( and there are other locales not supported).

### Main modification
- Get a language coverage file for DejaVu fonts from https://github.com/dejavu-fonts/dejavu-fonts/blob/version_2_34/dejavu-fonts/langcover.txt
- Parse that with a simple awk script, generate property file contains information about DejaVu fonts supporting the language and country code.
- On `DarkMonkeyLookAndFeel.java`, apply DejaVu font if it supports glyphs for current user's locale and country 

### Gradle
- Add a new task `copyEngineLibs` that copy dependency jar files under `./engine` directory into `lib` directory.
  - This task retrieve `corelibs, optlibs, testdatalibs, examplelibs` jar files name starting with `jme3-`, and renaming.
  - Exceptionally, `com.github.nifty-gui:nifty:1.4.2` jar will be copied.

- Remove `dependencies.resolve()` lines 
  - This line requires gradle to get dependency jars from maven repository
  - If there is no dependency jar files in local lib directory, this line causes error 
  - Current PR retrieve jar files with `copyEngineLibs` instead of `dependencies.resolve()`

```groovy
// example
copyBaseLibs.inputs.files configurations.corelibs.resolve()
copyBaseLibs.outputs.dir "jme3-core-baselibs/release/modules/ext/"
copyBaseLibs.outputs.dir "jme3-core-libraries/release/modules/ext/"
```

- misc
  - I compiled sdk with Java8
  - If project is compiled by Java11, the library javax.annotation.api will be required


jme3-welcome-screen/src/com/jme3/gde/welcome/WelcomeScreenTopComponent.java
```diff
@@ -57,7 +56,7 @@ public WelcomeScreenTopComponent() {
 
         jScrollPane2.setOpaque(false);
         jScrollPane2.getViewport().setOpaque(false);
-        jEditorPane1.putClientProperty(SwingUtilities2.AA_TEXT_PROPERTY_KEY, SwingUtilities2.AATextInfo.getAATextInfo(true));
```
MeFisto94, line 60: Looks good, but usually having separate PRs is wanted, it has a few pros.

jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/DarkMonkeyLookAndFeel.java
```diff
@@ -395,4 +401,29 @@ public String getThemeProperty(String name) {
...
+        if (!userLangIsCJK()) {
+            theme.setFont(DarkMonkeyLookAndFeel.DEJA_VU_SANS_FONT);
+        }
+    }
+
+    /**
+     * DejaVu doesn't support CJK glyphs
+     * @return true if user locale is "ja" or "ko" or "zh"
+     */
+    private Boolean userLangIsCJK() {
```
MeFisto94, line 415: How is this for glyphs from thailand or india, I guess they would have similar problems?

jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/DarkMonkeyLookAndFeel.java
```diff
@@ -395,4 +401,29 @@ public String getThemeProperty(String name) {
...
+        //
+        // >ja Japanese (0/6537) (0/6537) (0/6537)
+        // >ko Korean (0/2443) (0/2443) (0/2443)
+        // >zh-cn Chinese (simplified) 0% (2/6765) 0% (2/6765) 0% (2/6765)
+        // >zh-hk Chinese Hong Kong Supplementary Character Set (0/2213) (0/2213) (0/2213)
+        // >zh-mo Chinese in Macau (0/2213) (0/2213) (0/2213)
+        // >zh-sg Chinese in Singapore 0% (2/6765) 0% (2/6765) 0% (2/6765)
+        // >zh-tw Chinese (traditional) (0/13063) (0/13063) (0/13063)
+        String lang = System.getProperty("user.language");
+        return lang.matches("ja|ko|zh");
```
MeFisto94, line 427: Is something like startsWith required for Chinese or maybe `^zh-.*$` or something?

jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/DarkMonkeyLookAndFeel.java
```diff
@@ -29,9 +29,10 @@
 public class DarkMonkeyLookAndFeel extends com.nilo.plaf.nimrod.NimRODLookAndFeel{
     
     public static final String dmLAFDefault = "DarkMonkey.theme";
+    public static final String DEJA_VU_SANS_CONDENSED_PLAIN_12 = "DejaVu Sans Condensed-PLAIN-12";
+    public static final Font DEJA_VU_SANS_FONT = Font.decode(DEJA_VU_SANS_CONDENSED_PLAIN_12);
```
MeFisto94, line 33: I don't know how expensive decoding a Font is, but if we don't need it (on CJK), we could move the Decoding into `applyDejaVuFontIfSupported`

build.gradle
```diff
@@ -16,13 +16,14 @@ if (!hasProperty('mainClass')) {
...
-        url "http://jcenter.bintray.com/"
+        url "https://jcenter.bintray.com/"
     }
+    mavenCentral()
     maven {
         url "http://updates.jmonkeyengine.org/maven/"
     }
     maven {
-        url 'http://nifty-gui.sourceforge.net/nifty-maven-repo/'
+        url "http://nifty-gui.sourceforge.net/nifty-maven-repo/"
```
MeFisto94, line 26: This change is okay

build.gradle
```diff
@@ -16,13 +16,14 @@ if (!hasProperty('mainClass')) {
 repositories {
     mavenLocal()
     jcenter {
-        url "http://jcenter.bintray.com/"
```
MeFisto94, line 19: This removes java 7 compat, while this is probably fine, you could also write it as `jcenter()` then instead.

jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/DarkMonkeyLookAndFeel.java
hangingman: Maybe so.
If so, I started to think it's better to find the list of what language is supported or not.
I will research what language is curretly supported by DejaVu font.

jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/DarkMonkeyLookAndFeel.java
hangingman: I don't think so. Because the language code and country code is different property.
For example my default java language code and country code setting are 
```
-Duser.language=ja -Duser.country=jp
```

UI should follow user's property of `user.language`.

jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/DarkMonkeyLookAndFeel.java
hangingman: I found list of supported language code.
https://github.com/dejavu-fonts/dejavu-fonts/blob/version_2_34/dejavu-fonts/langcover.txt

jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/DarkMonkeyLookAndFeel.java
hangingman: I changed my mind, there are some minor patterns (can't detect supported or not by only language code).
It looks good to detect with both language code and country code.

```
mn-cn  Mongolian in China                            (0/130)            (0/130)            (0/130)      
mn-mn  Mongolian in Mongolia                    100% (70/70)       100% (70/70)       100% (70/70)      
```

build.gradle
hangingman: Change to `jcenter()`

jme3-welcome-screen/src/com/jme3/gde/welcome/WelcomeScreenTopComponent.java
hangingman: I will create another PR for this.

jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/DarkMonkeyLookAndFeel.java
hangingman: I have created a list with property file with using this langcover.txt.

jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/DarkMonkeyLookAndFeel.java
hangingman: Done

hangingman:
@MeFisto94 Hi,
I updated this PR. Would you please review it again ?
If PR is OK, I will put together a commit.

build.gradle
```diff
@@ -15,14 +15,13 @@ if (!hasProperty('mainClass')) {
 
 repositories {
     mavenLocal()
-    jcenter {
-        url "http://jcenter.bintray.com/"
-    }
+    jcenter()
+    mavenCentral()
```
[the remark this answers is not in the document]
hangingman: Removed `mavenCentral()`

copy_libs.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
hangingman: I update build.gradle to copy jar files from engine directory, and removed shell script.

hangingman:
@MeFisto94 Hi, 

>Regarding the files I am not sure what the best approach would be, though, e.g. instead of awk having a gradle task auto-generate the file so it does not need to be in the repo? or not have that because the source file langcover.txt would never change.
Talking of langcover, does it have any license issues we should be aware of?

I think langcover.txt may not be changed in the future, so removed.

>I am still not sure about the existence of this, gradle's buildSdk task should take care of that in a better (version- and platform independent manner).

I added copyEngineLibs task in the build.gradle. This task copy dependency jar files. And described detailed description on issue top.

build.gradle
```diff
@@ -138,7 +140,59 @@ task checkPlatformConfig {
...
+    description = "Copy libraries from engine directory before configuring gradle"
+
+    doFirst {
+        // Collect dependencies in corelibs, optlibs, testdatalibs...
+        def allDependencies = project.configurations.collect { conf -> conf.incoming.dependencies }.flatten()
+        allDependencies.each { file -> copyEngineJarsToLib(file) }
+    }
+}
+
+def copyEngineJarsToLib(file, extraJars=["nifty"]) {
```
MeFisto94, line 153: I still don't gather why you try to copy the files from the engine folder, where the idea is to use gradles dependency resolution?

Also why do you depend on nifty-gui through jitpack while we get it transitively through https://github.com/jMonkeyEngine/sdk/blob/36d11b6f712539693e34bbc79009121c11aa1d53/build.gradle#L99 ?

hangingman: @MeFisto94 
Because, in my understanding, the `buildSdk` task need to have dependency jars but there are no process to download or copy these jar files. Probably since that reason,  on my local environment after executing `buildSdk` will cause gradle error "Could not find org.jmonkeyengine:jme3-*".

I'm afraid current build script don't resolve dependency jars (jme3-*.jar) correctly. `build_engine.sh` generates many jme3-*.jar files, but it don't copy files into `/lib` directory or maven local repository.
So, I modified build.gradle script. Is there any misunderstanding for my side ?

MeFisto94: That's definitely a misunderstanding, because CI works (https://travis-ci.org/github/jMonkeyEngine/sdk/jobs/724436054).
If that error was about a specific jme3-jogl, that's fixed in 56cdc94e55a18b4b05bf8e77467db08802c6d491, essentially: Your engine folder needs to be on 3.3 (checkout or delete it and let the scripts work).

Due to https://github.com/jMonkeyEngine/sdk/blob/36d11b6f712539693e34bbc79009121c11aa1d53/settings.gradle#L2, engine dependencies should be there (unless jogl, which got deleted by the engine)

https://github.com/jMonkeyEngine/sdk/blob/36d11b6f712539693e34bbc79009121c11aa1d53/build.gradle#L167 Is the part which is copying them into a netbeans module, from which it is then available.

So all in all, it should work. Sorry for being not clear enough about this in the beginning, so you had that extra work

hangingman:
Close PR, and will reopen after fixed.

Issue #252, opened 2020-03-13
Re-Applying changes from gltf files broken


MeFisto94:
See [here](https://hub.jmonkeyengine.org/t/solved-animation-not-working-in-sdk-scene-composer/42916) for context.

![img](https://jme-hub-cdn-jmonkeyengineor.netdna-ssl.com/uploads/default/original/4X/0/5/b/05bfb037a820d708da388e6ba928c9f37a37c9ca.png)

MeFisto94:
The cause is that it tries to use 3.2 SkeletonControls instead of Armature.

neph1:
Can this be closed?

MeFisto94:
I am not sure, I haven't really experimented with all possible side effects of re-applying gltfs again.
Did you spot something in the context of #315 ?
If not, we can probably close it and re-open an issue if that's required :)

PR #282, opened 2020-09-01
include libraries: Heart v6.0.0, Minie v3.0.0, Wes v0.6.0

stephengold:


MeFisto94:
Thanks for caring about the SDK! :)

stephengold:
I'm curious: what are your intentions regarding the SDK? What should I tell users and potential users?

MeFisto94:
I really want to release a 3.3 version, but there are a few fundamental changes that haven't been really tested and I'd want to move to a working update center again, so that's a bit "critical" to just release, which is why I'd need testing, but there aren't many available that would care about testing.

For instance, I've hacked 3.3 Animation System support in, but we have yet to see if there are more bugs than only #252. 
And the update system together with maybe cutting out modules like nifty to be not part of the standard distribution, requires even more testing, of course. But it could help us with future updates.

Maybe I should just do a testing build and see what happens

stephengold:
If there's a build you want tested, I'll install and test it. However, I might not test it very thoroughly.

PR #287, opened 2020-10-10
Update the updater splash image and set the branding colors

tonihele:
Updates the updater splash image to use the images supplied earlier. Also changed the branding colors, but I really didn't test these. Seems that the old values were the default ones on the branding.


tonihele:
I tested it, looks good!

PR #289, opened 2020-10-25
Projectfactory

tonihele:
- Tidied up some related files
- Use annotations instead of the meta-inf (I understood this is the newer way, seems smarter to me, everything else is already with annotations)
- Implement ProjectFactory2 interface instead of the ProjectFactory

The newer interface speeds up things quite a bit. Now the project wont be loaded for just the preview, instead it is "instantly" recognized. In reality I doubt people have much of these projects laying around to see any speedup :) But hey, now this is up-to-date and ready for someone to copy paste...

I didn't find any place where the normal JME project is actually loaded up. I can see in the open projects dialog that it is recognized. When it is actually loaded, the project view shows the standard java icon instead of the JME icon seen in the open projects dialog.... The asset pack & codeless projects do have the icon even in project view.

### DIFF
--- a/jme3-assetpack-support/src/META-INF/services/org.netbeans.spi.project.ProjectFactory
+++ b/jme3-assetpack-support/src/META-INF/services/org.netbeans.spi.project.ProjectFactory
@@ -1,1 +1,0 @@
-com.jme3.gde.assetpack.project.AssetPackProjectFactory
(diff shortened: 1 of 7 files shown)


MeFisto94:
Given that I'll wait until you have looked at the regular projecrt as well.
Take a look at the gradle PR, I think I touched something there, but I also remember searching for it. I think it's called something like JavaProject or something, so something which we would not expect.

tonihele:
I can take a look...

tonihele:
Hmm, I looked at the regular project as well. It is completely different case. Regular project is an extension, or rather the many types we actually extend with the asset manager. https://netbeans.apache.org/tutorials/nbm-projectextension.html. That part works I guess as expected. Only I am not sure where it actually picks up the icon to the open projects dialog but as far as I could see, it doesn't yet open the project itself. Only when you click the project in the dialog.

These changes are for the two new types defined: https://netbeans.apache.org/tutorials/nbm-projecttype.html. To prevent them to be opened just by seeing them somewhere.

I think this is fine as it is?

Issue #264, opened 2020-04-05
Remove OgreXML


MeFisto94:
Maybe as an intermediate solution add a warning dialogue to recommend gltf.

tonihele:
Also does this issue justify removing the million years old bundled Blender from the SDK? I don't see why it would be needed anymore. Was it that the particular version we use just happens to for with the OgleXML stuff? Now that it will be gone. We could remove Blender altogether?

The old version probably can't do GLTF. And I don't see much point in getting a newer version to bloat the SDK side. People would be free to choose their modeling software, and version. Am I right @MeFisto94 ?

MeFisto94:
The reason that we did bundle blender is twofold:
Yes, it is because that used to be some very specific version that was the only compatible with the ogre xml script I think, but we bumped that to 2.79, the last version that the .blend importer supported.

We also use it to invoke scripts on it to be able to import fbx without relying on the engines FBX importer (and maybe other formats such as .dae), which is exactly the same way unity does it, minus the fact that they search Unity on PATH and export to fbx and import fbx manually. We'd need to change that to support more recent blenders and import .dae, .fbx, .stl and what not into blender -> auto convert to gltf and import the gltf into j3o.

The only concern I have that way, is that we loose the ability to support non-standard-pbr-shading materials, because that's the only thing gltf can really express without extensions. But I guess if you're not willing to work in an age old blender version as an artist, that's the only way you got anyway.

neph1:
Assuming this complete

PR #309, opened 2022-03-19
#264 Removal of ogrexml. Unsure of whether I've caught everything, bu…

neph1:
…t it builds and loads. #264 
I haven't tested this extensively, just checked that the SDK  builds and loads. Anything particular?


MeFisto94:
I guess I am also targetting it for 3.5, which should be available "soon" :D (As soon as 3.4 is working, that is)

tonihele:
@MeFisto94  care to check this one?

MeFisto94 (commented):
Looks good, the only thing I am uncertain is/are the asset packs then.
Do we need to remove them, if there is no real use for them anymore?
Like isn't this world forge the only real implementation/use case?

neph1:
I thought AssetPacks were a generic implementation to distribute collections of assets? Just that noone ever used them.

MeFisto94:
True, I guess it's just the other way around: world forge was one of the reasons we had ogre import, I guess.
and yeah, makes sense to keep asset packs, maybe someday someone finds them useful

Issue #206, opened 2019-09-06
DirectionalLight gizmo in SceneComposer doesn't rotate around the correct axis


grizeldi:
No matter whether you try to rotate it with the rotation gizmo or just by pressing R, it spins around a seemingly random axis, not the one it's supposed to.

PR #310, opened 2022-03-20
fixes annoying directionallight rotation

neph1:
gizmo still flickers a bit, but at least rotation is consistent. #206


tonihele:
Nice, works way better! Thanks!

PR #311, opened 2022-03-20
Misc fixes

tonihele:
I seemed to have some changes done awhile ago. Might as well push them. Just some cleaning up. Switches the use of timer to something hopefully better. Seems to work and I think it is appropriate to use the NB common thread pool. The Audio device pausing message is one that pops up for me and that I used to test this.

https://netbeans.apache.org/wiki/DevFaqRequestProcessor.asciidoc


MeFisto94:
Actually I am not sure if the Default Processor is appropriate, because we never know if like every asset in the project could throw one notification. At the same time, we never need more than one notification at the same time.
Since we're talking one constructor (`new RequestProcessor("Notification processor", 1);`), I guess that's a very well idea,
It's far more explicit in dumps and reduces deadlocking risk

MeFisto94:
Actually I am not sure if the Default Processor is appropriate, because we never know if like every asset in the project could throw one notification. At the same time, we never need more than one notification at the same time.
Since we're talking one constructor (`new RequestProcessor("Notification processor", 1);`), I guess that's a very well idea,
It's far more explicit in dumps and reduces deadlocking risk

tonihele:
I updated it to use own request processor

MeFisto94:
sorry, forgot to toggle the squash button, but now can't easily squash and force-push, I guess we shouldn't do that.
Up side: it documents how annoying codacy is :D

PR #312, opened 2022-03-20
icon for gltf

neph1:
Sorry for spamming PR's, but it's always bothered me that the gltf extension uses the blender icon.


tonihele:
I would say that keep them coming. It is super cool to see some activity in the SDK project, big or small. And good to keep the PRs per feature, easy to check and integrate.

PR #313, opened 2022-03-20
grid and scene snapping

neph1:
A feature which for some reason was only half implemented(?). I know I have wished for this many times.

[video showing snapping in SDK](https://www.youtube.com/watch?v=gTGdQsGZoYY)


neph1:
crap. i committed a file which shouldn't have been. will update

neph1:
Interesting how it allowed the first commit through. Does it choose a couple of things at random to complain about? :)

neph1:
I'll see if I can resolve the complexity issue tonight. Might be a good thing so the snapping can be applied to other transforms in the future.

neph1:
Finally. Didn't want to autoformat, because that brought so many more "errors". Better to refactor that separately. Feel free to squash commit this.

neph1:
There are also two types of scene snapping (that I can think of). I guess the one I implemented is the more commonly requested, "snap to ground". The other would be "mouse pick snap" where you snap to whatever you click on. Most of that is already implemented in the PickManager class, so it could easily be added as well.

neph1:
Realized there's some work left; I'll change so that snapping only applies to the axis/axes you're editing. I'll also add 15 degree rotation snapping if grid snapping is enabled (unless there are complaints). Not 100% intuitive, but a useful feature, I think.

tonihele:
Issues
======
+ Solved 3
- Added 6

Heh, Codacy <3 Thoughts and prayer...

neph1:
(╯°□°)╯︵ ┻━┻

tonihele:
It would be so good to have static code analysis that can be also used in the IDE, with the same ruleset...

tonihele:
Very nice, thank you for your contribution!

Issue #314, opened 2022-03-23
j3o caching in scenecomposer


neph1:
User ([kevinba99](https://hub.jmonkeyengine.org/u/kevinba99)) report:

I asked this question before and I didn’t get any real help. But I’m noticing that the SDK and J30 scene editor is caching the file and editing the cached file and not the actual file inside the SDK (netbeans).
I can’t figure out how to get rid of caching and done some searches and nothing has helped.

file snake.glb → conver to j3o.
I open it and rename a node to “snake”, “body”, “eyes”.
Save it.

Go back into blender, make a change, changed color to black on the snake.
Go back into SDK, it asks if I want to update it. I say yes. then open the scene editor and the snake still brown.
I view it in a object viewer outside of the SDK and it is black.

I delete the j3o file.
Go back into SDK and convert model to j3o again and reopen it. It is till brown. and the nodes are still renamed even though the glb file has obj1 obj2 obj3 for names.

I then delete the j3o again. Then in file manager rename the snake.glb to snake1.glb and go back into SDK and conver snake1.glb to j3o.
Open snak1.j3o in scene editor and now the snake is black, and the nodes are obj1, obj2 and obj3.

Then go to file manager and rename snake1.* (blg, j3o and j3odata) back to snake.*.
Go back into SDK and open snake.j3o in scene editor and the snake is brown again and node name are still renamed, even though in the j3o file they are not.

[https://hub.jmonkeyengine.org/t/j3o-scene-and-sdk/45421](https://hub.jmonkeyengine.org/t/j3o-scene-and-sdk/45421)

neph1:
This isn't actually a bug, merely an unexpected outcome. The message says it will update the mesh data, and it does, even if the object is open in the SceneComposer. It even has options for animations. It doesn't however update anytyhing else, like materials or names of the objects.

neph1:
I propose to expand this to include more data, including materials.

MeFisto94:
Just one note here, that there may indeed be a cache, but we're talking stock jme there:
AssetManager has a weak-key cache on assets, so even if you were to re-load the whole asset by string (and we don't have any other caching in place, that is), you'd end up with a cached version.

neph1:
I looked at the caching, and it seems to be used and closed. Updating the spatial really works fine, it's just that it doesn't do more than update the mesh. In my case I think I've been misled by that and the fact that it didn't use AnimComposer.

neph1:
[...]

PR #315, opened 2022-03-23
updates material. null check if mesh names have changed

neph1:
Proposed fix for #314


MeFisto94:
Looking good to me, but I think we should use Material#clone.
Worst case, we produce some garbage overhead.
I come to that thinking because of, as said in #314, it may be that trying to load the j3o (or even blend/gltf!) again, may have the same Material instance, and if we share a mutable reference, there may be issues where having multiple scene explorers open you can trigger undesired side effects.

Besides that, I have the nit of adding a few more spaces (e.g between `//` and the actual comment and around the if statement)
Also taking the chance on thanking you again for your big efforts on helping out on the SDK :)

neph1:
I agree on using clone. I came to the same conclusion while thinking about the PR. Will amend tonight

Np. Might as well thank you for keeping the SDK alive for so long :)

neph1:
"First sentence of Javadoc is missing an ending period." omfg! Squash please.

MeFisto94:
True, I guess that's what your other PR is set for to do :)

Issue #83, opened 2016-07-04
[Netbeans Node API] Nodes don't refresh recursively


MeFisto94:
So Netbeans has the Nodes API to display it's hiearchic structures like you see the SceneGraph in the Scene Explorer.

Now we have that "AbstractSceneExplorerNode" which is the Base-Class to many Nodes.
It features a refresh method which contains the following:
```java
//TODO: refresh does not work
public void refresh(boolean immediate) {
   if (jmeChildren instanceof JmeSpatialChildren) {
      ((JmeSpatialChildren) jmeChildren).refreshChildren(immediate);
   }
}
```

The Problem here actually is, that those jmeChildren itself don't seem to trigger it's nodes.
I tried to add the following code, but for some reason, the children don't seem to have any nodes attached to. I don't have the time to work into the Nodes API again, though.

Maybe you can read everything in the Commits regarding the MotionEvent Implementation, maybe those update as they should. They also use the Parent/Children Stuff, but it's not JmeSpatialChildren, of course.

```java
for (Node n : nodes) { // TODO: Why is nodes always empty? What about the Children's Children?
  if (n instanceof AbstractSceneExplorerNode) {
     ((AbstractSceneExplorerNode)n).refresh(immediate);
  }
}
```
(JmeSpatialChildren:refreshChildren)

Ohh, and the AnimationControl Stuff has the same API regarding Children and Nodes

neph1:
Is this still relevant? I can't remember seeing any issue regarding this. How to reproduce?

PR #316, opened 2022-03-25
Animcomposer and external change refactor

neph1:
I'd like some opinions on this. Primary goal is to replace AnimComposer when updating the gltf, which it seems to do, although you need to reopen the .j3o to see it.

But mostly I'd like to know your thoughts on the refactoring. I'm thinking it can be taken further and maybe have a selectable list of transfers the user might want to do, rather than automating everything. It's out of scope for now though.

### DIFF
--- a/jme3-codepalette/src/com/jme3/gde/codepalette/layer.xml
+++ b/jme3-codepalette/src/com/jme3/gde/codepalette/layer.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE filesystem PUBLIC "-//NetBeans//DTD Filesystem 1.1//EN" "http://www.netbeans.org/dtds/filesystem-1_1.dtd">
+<!DOCTYPE filesystem PUBLIC "-//NetBeans//DTD Filesystem 1.2//EN" "http://www.netbeans.org/dtds/filesystem-1_2.dtd">
 <filesystem>
     <folder name="JavaPalette">
         <folder name="JME Material">
--- a/jme3-core/src/com/jme3/gde/core/util/SpatialUtil.java
+++ b/jme3-core/src/com/jme3/gde/core/util/SpatialUtil.java
@@ -50,6 +50,7 @@ public class SpatialUtil {
 
     public static final String ORIGINAL_NAME = "ORIGINAL_NAME";
     public static final String ORIGINAL_PATH = "ORIGINAL_PATH";
+    
     private static final Logger LOGGER =
             Logger.getLogger(SpatialUtil.class.getName());
 
--- a/jme3-core/src/com/jme3/gde/core/util/SpatialUtil.java
+++ b/jme3-core/src/com/jme3/gde/core/util/SpatialUtil.java
@@ -50,7 +50,7 @@ public class SpatialUtil {
 
     public static final String ORIGINAL_NAME = "ORIGINAL_NAME";
     public static final String ORIGINAL_PATH = "ORIGINAL_PATH";
-    private static final Logger logger =
+    private static final Logger LOGGER =
             Logger.getLogger(SpatialUtil.class.getName());
 
     /**
@@ -66,7 +66,7 @@ public class SpatialUtil {
         while (spat != null) {
             String name = spat.getName();
             if (name == null) {
-                logger.log(Level.WARNING, "Null spatial name!");
+                LOGGER.log(Level.WARNING, "Null spatial name!");
                 name = "null";
             }
             geometryIdentifier.insert(0, name);
@@ -90,24 +90,23 @@ public class SpatialUtil {
                 Spatial curSpat = geom;
                 String geomName = geom.getName();
                 if (geomName == null) {
-                    logger.log(Level.WARNING, "Null Spatial name!");
+                    LOGGER.log(Level.WARNING, "Null Spatial name!");
                     geomName = "null";
                 }
                 geom.setUserData("ORIGINAL_NAME", geomName);
-                logger.log(Level.FINE, "Set ORIGINAL_NAME for {0}",
+                LOGGER.log(Level.FINE, "Set ORIGINAL_NAME for {0}",
                         geomName);
                 String id = SpatialUtil.getSpatialPath(curSpat);
                 if (geomMap.contains(id)) {
-                    logger.log(Level.WARNING, "Cannot create unique name "
-                            + "for Spatial {0}: {1}", new Object[]{geom,
-                            id});
+                    LOGGER.log(Level.WARNING, "Cannot create unique name "
+                            + "for Spatial {0}: {1}", new Object[]{geom, id});
                 }
                 geomMap.add(id);
                 geom.setUserData("ORIGINAL_PATH", id);
-                logger.log(Level.FINE, "Set ORIGINAL_PATH for {0}", id);
+                LOGGER.log(Level.FINE, "Set ORIGINAL_PATH for {0}", id);
             });
         } else {
-            logger.log(Level.SEVERE, "No Spatial available when trying to add"
+            LOGGER.log(Level.SEVERE, "No Spatial available when trying to add"
                     + " Spatial paths.");
         }
     }
--- a/jme3-core/src/com/jme3/gde/core/util/SpatialUtil.java
+++ b/jme3-core/src/com/jme3/gde/core/util/SpatialUtil.java
@@ -32,7 +32,6 @@
 package com.jme3.gde.core.util;
 
 import com.jme3.anim.AnimComposer;
-import com.jme3.scene.SceneGraphVisitor;
 import com.jme3.scene.Spatial;
 
 import java.util.ArrayList;
@@ -47,20 +46,20 @@ import java.util.logging.Logger;
  *
  * @author normenhansen
  */
-@SuppressWarnings({"unchecked", "rawtypes"})
 public class SpatialUtil {
 
+    public static final String ORIGINAL_NAME = "ORIGINAL_NAME";
+    public static final String ORIGINAL_PATH = "ORIGINAL_PATH";
     private static final Logger logger =
             Logger.getLogger(SpatialUtil.class.getName());
-
 
     /**
      * Gets a "pathname" for the given Spatial, combines the Spatials and
      * parents names to make a long name. This "path" is stored in geometry
      * after the first import for example.
      *
-     * @param spat
-     * @return
+     * @param spat Spatial
+     * @return id of spatial
      */
     public static String getSpatialPath(Spatial spat) {
         StringBuilder geometryIdentifier = new StringBuilder();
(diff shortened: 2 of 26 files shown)


neph1:
Since I don't know how to make a checklist option, I added an option to either apply all data or only the mesh data (or none). Should satisfy most needs. I don't like that it loses the option to select whether to only apply mesh and animation data (like before)

neph1:
Not sure what happened here, I must have missed adding a couple of files before committing, then made a hard reset of the branch because they were missing locally too. The dangers of context switching I guess.
I think this is done feature wise. Not sure I'll be able to fix all those codacy issues, though.

jme3-codepalette/src/com/jme3/gde/codepalette/layer.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE filesystem PUBLIC "-//NetBeans//DTD Filesystem 1.2//EN" "http://www.netbeans.org/dtds/filesystem-1_2.dtd">
```
tonihele, line 2: Unwanted merge issue?

tonihele: Same for the .keep?

neph1: It would seem so. I did have some issues with the branch, previously.

neph1:
Pause this for a bit. I found an issue while testing.

neph1:
I hadn't checked before that animations actually _worked_, just that they had been updated. I've fixed that now, however, they don't work in the scenecomposer (without reloading the j3o). My guess is that the JmeAnimComposer needs to be updated with the cloned AnimComposer. Since you're the author, @MeFisto94, is there a convenient method of doing such things?

MeFisto94:
I don't remember _too much_ of it, just that the `JmeXyz` are counterpart to engine `Xyz` classes and basically are mostly there for having Java Bean Properties and a tree structure. We had this `refreshChildren` thing on JmeNodes or something, but I don't know if it fixes that. Technically, the cloned AnimComposer is a new AnimComposer that should have a new JmeAnimComposer that is then put into this tree hierarchy and refreshed, I think.

So instead of trying to change stuff in there, I guess the official way would be to kinda reload the node tree in some way (but I don't remember API specifics there)

neph1:
I did look for a way to refresh the node tree. I assumed there would be a
similar problem with other things, but didn't find it. I'll give it another
try, though.

MeFisto94:
Looking at #83 this should be possible.
Something like AbstractSceneExplorerNode#refresh. And ideally those node classes extend that.
If that doesn't help, give me a ping and I do a more thorough search

MeFisto94:
I'd love if someone would help in testing this PR :) I'll still need to review it, but since it's rather large, I probably need to schedule some time for it

neph1:
It may not be as big as it looks. Most of the code for Mesh and Animation transfer is taken from SpatialUtil, only broken out into separate classes, and Material and Transform transfer is built on the same "pipeline".
I've tested everything quite well, but since the Animation transfer is newest, and I was a bit exhausted with the whole thing, there may be edge-cases I didn't cover.
Maybe I should do another documentation commit.

jme3-core/src/com/jme3/gde/core/util/TaggedSpatialFinder.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.jme3.scene.Spatial;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Finds a previously marked spatial in the supplied root Spatial, creates
+ * the name and path to be looked for from the given needle Spatial.
+ */
+public class TaggedSpatialFinder {
```
neph1, line 11: This is broken out from SpatialUtil

jme3-core/src/com/jme3/gde/core/util/datatransfer/MeshDataFromOriginal.java
```diff
@@ -0,0 +1,96 @@
+package com.jme3.gde.core.util.datatransfer;
```
neph1, line 1: I think this whole class is using code from SpatialUtil

neph1:
@MeFisto94 If you want to, we could try scheduling an hour and do the review together. Regarding testing, perhaps we can push out a beta to let others try it.

jme3-core/src/com/jme3/gde/core/util/datatransfer/TransformDataFromOriginal.java
```diff
@@ -0,0 +1,55 @@
...
+import com.jme3.scene.SceneGraphVisitorAdapter;
+import com.jme3.scene.Spatial;
+
+import java.util.logging.Logger;
+
+/**
+ * Copies Transform data (translation, rotation, scale) from an updated spatial
+ * to the original.
+ */
+public final class TransformDataFromOriginal implements
```
MeFisto94, line 16: Minor: The name is a bit confusing, as it can be read like "transform this data from original", something like "TransferTransformDataFromOriginal" maybe? "CopyTransformDataFromOriginal"?

jme3-core/src/com/jme3/gde/core/util/datatransfer/TransformDataFromOriginal.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public void update(final Spatial root, final Spatial original) {
+        original.depthFirstTraversal(new SceneGraphVisitorAdapter() {
+
+            @Override
+            public void visit(final Geometry geom) {
+                final Geometry spat = (Geometry) finder.find(root, geom);
+                if (spat != null) {
+                    spat.setLocalTransform(geom.getLocalTransform());
+                    LOGGER.log(ApplicationLogHandler.LogLevel.FINE,
```
MeFisto94, line 37: Is this LogLevel import the right one?

jme3-core/src/com/jme3/gde/core/util/datatransfer/MeshDataFromOriginal.java
```diff
@@ -0,0 +1,96 @@
...
+import com.jme3.scene.Node;
+import com.jme3.scene.SceneGraphVisitorAdapter;
+import com.jme3.scene.Spatial;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Copies mesh data from an updated external file to the spatial
+ */
+public class MeshDataFromOriginal implements SpatialDataTransferInterface {
```
MeFisto94, line 16: this name is not misleading (see the comment on TransformDataFromOriginal), but if you change it there, you also need to change it here, I guess.

jme3-core/src/com/jme3/gde/core/util/datatransfer/MeshDataFromOriginal.java
```diff
@@ -0,0 +1,96 @@
+package com.jme3.gde.core.util.datatransfer;
+
+import com.jme3.gde.core.scene.ApplicationLogHandler.LogLevel;
```
MeFisto94, line 3: Wrong import

jme3-core/src/com/jme3/gde/core/util/datatransfer/AnimationDataFromOriginal.java
```diff
@@ -0,0 +1,121 @@
...
+        original.depthFirstTraversal(new SceneGraphVisitor() {
+
+            @Override
+            public void visit(final Spatial spatial) {
+                final AnimComposer animComposer
+                        = spatial.getControl(AnimComposer.class);
+                if (animComposer != null) {
+                    final Spatial mySpatial = finder.find(root, spatial);
+                    if (mySpatial != null) {
+                        //TODO: move attachments: have to scan through all
```
MeFisto94, line 44: Should we create an issue for that and maybe mention that in the TODO comment and vice versa?

jme3-core/src/com/jme3/gde/core/util/TaggedSpatialFinder.java
```diff
@@ -0,0 +1,68 @@
...
+    
+    private static final Logger LOGGER =
+            Logger.getLogger(TaggedSpatialFinder.class.getName());
+    
+    public TaggedSpatialFinder() {
```
MeFisto94, line 16: We could honor this Codacy opinion, an empty default cosntructor is not gaining us anything

jme3-core/src/com/jme3/gde/core/util/TaggedSpatialFinder.java
```diff
@@ -0,0 +1,68 @@
...
+                } else {
+                    LOGGER.log(Level.WARNING, "Found Spatial {0} twice in"
+                            + " {1}", new Object[]{path, root});
+                }
+            }
+        });
+        return holder.spatial;
+    }
+
+    private static class SpatialHolder {
```
MeFisto94, line 65: This could probably have been an AtomicReference<Spatial>, but this is well enough, especially if it was code you only copied

jme3-core/src/com/jme3/gde/core/util/SpatialUtil.java
```diff
@@ -31,14 +31,9 @@
  */
```
MeFisto94, line 31: Nice refactorings in this class!

MeFisto94 (approved):
Looking good, there are just a few things that need fixing and those mostly apply to all classes from the datatransfer packet: Wrong LogLevel Imports and maybe considering giving them a different name. They may not come from your PR directly, but I guess we can fix them easily anyway

jme3-core/src/com/jme3/gde/core/util/datatransfer/MeshDataFromOriginal.java
neph1: Consistency ftw!

jme3-core/src/com/jme3/gde/core/util/datatransfer/TransformDataFromOriginal.java
neph1: Not sure. It's used conservatively and I probably just copied this from SpatialUtil. That log is used to display the info bubbles and suck. But I think the SDK is too chatty with these things anyway, so it makes sense to remove it.

jme3-core/src/com/jme3/gde/core/util/datatransfer/AnimationDataFromOriginal.java
neph1: Yes, that is an old comment, but I think it's still valid. But it's probably a rare use case.

jme3-core/src/com/jme3/gde/core/util/TaggedSpatialFinder.java
neph1: I think I added it because Codacy complained there was no default constructor..

PR #317, opened 2022-03-26
adding "missing" folder

neph1:
This has annoyed me for years.


MeFisto94:
I guess we need to ensure to also have this for the current and coming gradle template at some point

PR #318, opened 2022-03-29
making the cursor arrow more usable, show position in panel.

neph1:
Might seem like a minor thing, but sometimes you want to position things in code and currently you need to create a new Node and drag it around until it's where you want it. Now you can just point wherever and see the position.

### DIFF
--- a/jme3-core/src/com/jme3/gde/core/scene/controller/SceneToolController.java
+++ b/jme3-core/src/com/jme3/gde/core/scene/controller/SceneToolController.java
@@ -67,11 +67,19 @@ public class SceneToolController extends AbstractAppState {
     protected AssetManager manager;
     protected Material blueMat;
     protected AbstractCameraController camController;
+    
     private SceneToolControllerListener toolListener;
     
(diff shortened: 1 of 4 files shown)


MeFisto94:
It feels we really need to do some taming on Codacy and sadly the inline comments it does make no sense at all

neph1:
Most of the issues here are due to the NB generated UI code.

tonihele:
Hmm, point how? I'm trying this out but it always displays:
image

neph1:
"Works on my machine"

Hmm. Just to make sure; You're not clicking on the void? I'll see if I can find out something

tonihele:
Here is how I tried it, did I misunderstood how this should work:
https://www.youtube.com/watch?v=SPYq5kRwqLw

tonihele:
Ah okay, it does work. Invalid user error. You are supposed to click on a mesh to position the cursor... Okay okay okay, it works :)

tonihele:
Thank you for the contribution!

PR #331, opened 2022-05-11
adding block statements to Technique to retain any added Defines

neph1:
Maybe something for 3.5

### DIFF
--- a/jme3-materialeditor/src/com/jme3/gde/materialdefinition/fileStructure/UberStatement.java
+++ b/jme3-materialeditor/src/com/jme3/gde/materialdefinition/fileStructure/UberStatement.java
@@ -13,7 +13,8 @@ import java.util.LinkedList;
 import java.util.List;
 
 /**
- * A statement that will be enclosed with curly braces
+ * A statement that will be enclosed with curly braces.
+ * 
  * @author Nehon
  */
 @SuppressWarnings({"unchecked", "rawtypes"})
--- a/jme3-materialeditor/src/com/jme3/gde/materialdefinition/fileStructure/UberStatement.java
+++ b/jme3-materialeditor/src/com/jme3/gde/materialdefinition/fileStructure/UberStatement.java
@@ -13,7 +13,7 @@ import java.util.LinkedList;
 import java.util.List;
 
 /**
- *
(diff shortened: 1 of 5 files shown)


neph1:
One might argue that defines don't belong in shader nodes, which i guess is true for pure shader nodes. I'm however working on a hybrid PBR model [ref here](https://hub.jmonkeyengine.org/t/question-about-shader-editor/45498/46) which uses them to define textures (like PBRLighting). Possibly this is unnecessary duplication, since there is a DefinitionsBlock used in other places.

neph1:
This is included in #334

jme3-materialeditor/src/com/jme3/gde/materialdefinition/fileStructure/UberStatement.java
```diff
@@ -1,62 +1,62 @@
...
+import com.jme3.util.blockparser.Statement;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * A statement that will be enclosed with curly braces
```
MeFisto94, line 16: Nice!

MeFisto94 (approved):
Looking good besides the whitespace changes.
Do you know why they happen (i.e. what is the Line Endings in the file? Or is it coming from tabs vs spaces?)

neph1:
> Looking good besides the whitespace changes. Do you know why they happen (i.e. what is the Line Endings in the file? Or is it coming from tabs vs spaces?)

I think it's line endings, why I don't know. I use "autocrlf = false". But mostly I don't know what to do about it. I could try "autocrlf = input" and create a new branch from master and see if it helps.

neph1:
Ok. I have this plugin in netbeans from before that shows the type of line ending. It seems a lot of files have windows encoding (clrf). I probably had git set to change to unix endings (lf) before. But I have changed that since I started noticing these issues and I just tried reverting Uberstatement to the one on the server and reapply the change I did. And now it retains the clrf ending. So I could redo those files that have the problems.

Issue #335, opened 2022-05-20
Shaders need to be declared in "correct" order


neph1:
Or the shader won't compile as it looks for connections. This instantly becomes a problem when adding more than the standard Frag shader since they end up after it in the def file. I guess some kind of mapping is needed afterwards.

neph1:
I imagine this can be done in two ways.
My first thought was to load all into a Map and then find the connections from there. There'd need to be checks for cyclic dependencies, though.
Maybe it's easier to make sure they're saved in the correct order. That could be done with a Comparator, I believe, moving them down in the list if there is a dependency.

PR #338, opened 2022-05-22
sorting for shadernodeblock.

neph1:
I think this puts the shader nodes in order, at least when edited. It won't fix everything retroactively. Should solve  #335

### DIFF
--- a/jme3-materialeditor/src/com/jme3/gde/materialdefinition/fileStructure/ShaderNodesBlock.java
+++ b/jme3-materialeditor/src/com/jme3/gde/materialdefinition/fileStructure/ShaderNodesBlock.java
@@ -14,14 +14,15 @@ import org.openide.util.WeakListeners;
 
 /**
  * A ShaderNodeBlock is a shader node in the MatDef file.
+ *
  * @author Nehon
  */
 public class ShaderNodesBlock extends UberStatement implements PropertyChangeListener {
-    
+
     protected ShaderNodesBlock(int lineNumber, String line) {
         super(lineNumber, line);
     }
-    
+
     public ShaderNodesBlock(Statement sta) {
         this(sta.getLineNumber(), sta.getLine());
         for (Statement statement : sta.getContents()) {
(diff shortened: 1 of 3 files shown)


neph1:
I just noticed that ShaderNodeBlock has a similar compareTo method, but I guess it doesn't work as intended.

neph1:
I'm continuing work on this. It doesn't consider ancestor nodes, so it can't move a node more than one step in either direction, still giving the error for long setups. Why the old compareTo didn't work I don't know, but it becomes redundant with the new sorting.

jme3-materialeditor/src/com/jme3/gde/materialdefinition/fileStructure/ShaderNodesBlock.java
```diff
@@ -29,31 +32,95 @@ public ShaderNodesBlock(Statement sta) {
...
+
     public boolean removeShaderNode(ShaderNodeBlock shaderNodeBlock) {
         return contents.remove(shaderNodeBlock);
     }
-    
+
+    @Override
     public void propertyChange(PropertyChangeEvent evt) {
-        if (evt.getPropertyName().equals("order")) {
+        if (evt.getPropertyName().equals(ShaderNodeBlock.ADD_MAPPING) || evt.getPropertyName().equals("order")) {
```
neph1, line 51: Adding a listen for ADD_MAPPING  here so that we don't get the bubble saying the node is undeclared when adding a connection to a node higher up in the hierachy

jme3-materialeditor/src/com/jme3/gde/materialdefinition/fileStructure/ShaderNodesBlock.java
```diff
@@ -29,31 +32,95 @@ public ShaderNodesBlock(Statement sta) {
...
+        }
     }
-    
+
     public void sort() {
         List<ShaderNodeBlock> list = getShaderNodes();
-        Collections.sort(list);        
+        int passes = 0;
+        NodeComparator nodeComparator = new NodeComparator(list);
+        while (nodeComparator.changes != 0 && passes < 10) {
```
neph1, line 60: "Noone will ever require more than 10 sorting passes"

neph1:
Two main changes here.
1. Recursive searching in the node hierarchy for input connections and multiple sorting passes
2. Listening for mapping changes so that ShaderNodeLoaderDelegate doesn't have to throw its exceptions.

neph1:
Anyone wants to review this? Or should I just merge it?

tonihele:
I think it is fine to merge. You have at least received two thumbs up and you are probably the one with the most domain knowledge on this :D So go ahead.

I can only say about the code in general that it looks to have a complexity issue in a quick glance. Goes through the list many time inside a loop. It sorts a list, and the said list is traversed multiple times while sorting. So I would say it has an exponential algorithmic complexity. That is not a good thing, but whether it is noticeable is beyond me. I depends how much stuff there is in this graph.

tonihele:
I don't know the reasons behind this sort or anything, but is it this problem: https://en.wikipedia.org/wiki/Topological_sorting?

neph1:
It's been so long now :D but yes, I think so. The order is not apparent by parsing the list once (unless building a complex mapping, I guess). I think one option would be to do the sorting when connecting nodes, and only involve the connected nodes. This is a lazier approach, but I didn't notice any issues.

tonihele:
Hmm, if it is that, it says that the problem can be solved with linear algorithm. This this exponential. This would be interesting thing to tweak on and fix the algorithm. Maybe I'll poke it at some point. Would be nice to make unit tests for it. Last time I tried to make unit tests to SDK project.... it was a bit tricky with the Ant build. Perhaps Maven built SDK would also enable this more easier, the unit test part.

neph1:
If I remember correctly, the main issue is that it only moves the items one
step per pass. A custom sorter would work as well, I think. So, yes,
definitely something that could be improved.

PR #341, opened 2022-05-23
fixes #339 Attribute nodes are expected to be named "Attr"

### LINKED ISSUE #339: Attribute nodes are expected to be named "Attr" in the MatDef connections.
Currently when you add an Attribute (which is called "Attribute" in the MatDef) and reload the shader you're treated with an ugly error dialog.

neph1:
This fix is includes in my behemoth PR as well, although refactoring makes it differ slightly. I figure it's better if I can produce a couple of smaller PR's, then merge them (and deal with the conflicts) before someone tackles the big PR.

### DIFF
--- a/jme3-materialeditor/src/com/jme3/gde/materialdefinition/editor/ShaderNodePanel.java
+++ b/jme3-materialeditor/src/com/jme3/gde/materialdefinition/editor/ShaderNodePanel.java
@@ -194,7 +194,7 @@ public class ShaderNodePanel extends NodePanel implements InOut,
                 break;
             case Attribute:
                 header.setIcon(Icons.attrib);
-                setNameAndTitle("Attribute"); // sets text _and_ tooltip the same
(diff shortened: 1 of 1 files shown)


MeFisto94:
If it's easier/acceptable for you to commit individual PRs (which makes sense, seeing they are distinct fixes), then I'd welcome that :)

neph1:
It's certainly not easier, but I know the equations of real life and it will hopefully make things merge quicker :)
Merge away. The plan is to keep merging master into my "main" PR branch and hopefully it will get smaller. There's one more feature I might be able to break out of it and then there's only previews/editors left in it. It will still be substantial, but the changes to the core/existing classes should be managable.

neph1:
A middle ground perhaps would be look through the smaller PR's and get an overview what is changed and then tackle the big one without merging them. That might save me a lot of merging head-aches. But lets try merging a few of the smaller ones and see how it goes.

Issue #175, opened 2018-11-07
Lightprobe process fails to complete quite often.


mitm001:
Been trying out light probes lately and I keep having an issue where the light probe process will not complete past 73%. 

Can hit refresh and the next process will complete, however, the original attachment process remains stuck at 73%.

Testing is on a 2mx2m cube with one pbr material so it shouldn't even have a problem.

grizeldi:
Never had a problem with that, but atm light probes are a bit of a mess in the sdk. The fact that their code changes with every engine release doesn't help either.

The actual issue here could do with multithreading, since light probes create the env maps in a separate thread and maybe that thread somehow crashes or just fails to update the indicators?

mitm001:
I am not sure at all what's going on since it just stops at 73% every time it happens. Would seem to be something other than a crash since the process requires killing to stop it. Maybe its stuck in an infinite loop.

neph1:
Editor says: "Generating environment maps. Prefiltered env map face 5 generated."

java.lang.IllegalArgumentException: Cannot decrease processed workunit count (2) to lower value than before (4)
	at org.netbeans.modules.progress.spi.InternalHandle.progress(InternalHandle.java:322)
	at org.netbeans.api.progress.ProgressHandle.progress(ProgressHandle.java:264)
	at org.netbeans.api.progress.ProgressHandle.progress(ProgressHandle.java:242)
	at com.jme3.gde.core.sceneexplorer.nodes.JmeLightProbeProgressHandler.progress(JmeLightProbeProgressHandler.java:58)
	at com.jme3.environment.LightProbeFactory$JobListener.progress(LightProbeFactory.java:299)
	at com.jme3.environment.generation.RunnableWithProgress.progress(RunnableWithProgress.java:78)
	at com.jme3.environment.generation.PrefilteredEnvMapFaceGenerator.generatePrefilteredEnvMap(PrefilteredEnvMapFaceGenerator.java:195)

neph1:
It doesn't seem to be a problem with the generation per se, rather the reporting. I have a potential fix for this, but will test a bit more.

PR #344, opened 2022-05-26
#175 lightprobe process fails

neph1:
More of a work around I guess. I found a lot of hits for the specific error in netbeans bug list itself, but no solution (links were too old). But switching to "indeterminate" (no percentage progress bar) seems to work. I've generated around 50 light probes without the issue appearing (used to happen within the first 3 before). I've also generated a couple of "high quality bakes" which take much longer.


neph1:
"Fixes" #175

PR #351, opened 2022-07-09
Get the latest JDK 11

tonihele:
Just a simple JDK update. Scripts still work. I use this version myself and it has been out for some time now.


tonihele:
I'll just merge, I think this is so trivial thing...

PR #354, opened 2022-07-22
New splash screen for 3.4

nickidebruyn:


tonihele:
Hmm, if it is all possible, I would prefer not to have the version number in the resources/splashscreen/updatersplash.gif. It didn't use to have one. Just to ease the maintenance burden :D

nickidebruyn:
Hi @tonihele , can you check if this is better?
I check in a fix for the updatersplash.

tonihele:
Awesome thanks! Let's merge after 24h to wait for any other comments

PR #357, opened 2022-08-03
NullpointerException on Mesh selection

nickidebruyn:


tonihele:
LGTM, verified fix. No such property

Issue #328, opened 2022-05-01
Upgrade to Netbeans 14


MeFisto94:

tonihele:
A side note, Netbeans will no longer have LTS releases. Netbeans 12 was the last of such.

For the SDK I think this means that we should just maybe get the latest NB for a SDK release.

tonihele:
Which is 14 now :)

tonihele:
Closed with https://github.com/jMonkeyEngine/sdk/pull/358

PR #358, opened 2022-08-04
Upgrade to Netbeans 14

tonihele:
It is almost time for Netbeans 15 but lets go with 14 at least for SDK 3.4


tonihele:
Resolves #328

PR #360, opened 2022-08-04
Removed reference to old jme-bullet.jar and point to jbullet

nickidebruyn:
When creating a BasicGame in the SDK with the Ant build system it was still referencing the removed jme-bullet.jar.
I changed it to rather point to the jbullet implementation of the physics engine.
Also updated the javac.source and javac.target to JAVA 11.


tonihele:
Looks good to me! Thanks!

PR #361, opened 2022-08-14
Upgrade the installer from Netbeans 14

tonihele:
https://issues.apache.org/jira/browse/NETBEANS-3342 being officially resolved some time ago... Upgrade these libraries from Netbeans 14 as per the instructions.


MeFisto94 (approved):
We need to see if that fixes problems with installing and or causes new ones (i.e. required jvm*), but we can only see that in the field

* = something is wrong with the jvms that come with the installer anyway. It feels like they're only used once the SDK has been installed, at best.

tonihele:
Yeah, I have the impression that java is needed to install SDK. And the shipped JDK is not used at this stage..

PR #378, opened 2022-09-10
Mark examplelibs as non-transitive, enable javadoc and sources for all

### LINKED ISSUE #283: Javadoc and Sources seem missing for the sdk's libraries
I.e. when a user is doing a project, they only have the method signatures, but that's not feasible.

### LINKED ISSUE #368: All dependencies with sources get copied to jME example project
All dependencies with sources get copied to jME example project. Just that I don't forget. Doesn't cause any issues now, I enabled only sources to be downloaded for the example project anyway...

Line looks correct but good 'ol println debugging showed it gets all, I don't know, dependencies of dependencies?
https://github.com/jMonkeyEngine/sdk/blob/master/build.gradle#L486

tonihele:
This fixes #368, fixes #283.

Mark the examplelibs configuration to only contain itself, since we actually ship the dependencies to the created project. This library is used only to copy its sources to the project template.

What do you think @MeFisto94 ?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,33 +28,37 @@ configurations {
     examplelibs
 }
 
+configurations.examplelibs {
+    transitive = false
+}
+
 dependencies {
 
     corelibs dep("com.github.xbuf:jme3_xbuf:0.9.1", false, false)
     corelibs dep("com.badlogicgames.gdx:gdx-ai:1.8.2", true, true)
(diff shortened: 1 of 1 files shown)


MeFisto94:
good catch with the transitive = false. The `false, false` were coming from me where we still manually built the engine and thus it was not working, but due to us being back on maven central releases, that works again. Nice

MeFisto94 (approved):

Issue #379, opened 2022-09-11
File is not part of a jME project but tries to find original model


peedeeboy:
As reported by @tonihele in #373:

> And then there is this, not maybe 100% related but: WARNING [com.jme3.gde.core.assets.ExternalChangeScanner]: File is not part of a jME project but tries to find original model...
> 
> If I create a scene file to the Gradle template... It doesn't seem to understand it is a jME project or something....

This only appears to affect SDK3.4 (and latest Dev build) on Windows, Linux seems ok to me

peedeeboy:
So, I got Scenes to open in the latest Dev build on Windows.  The problem lies in `AssestLookupProvider.java`.  Line 106:
```java
if (gradle.getSubProjects().containsKey("assets"))
```
On my Linux machine, `getSubProjects()` returns the name of the Assets sub-project as `assets`, but on my Windows box, it gets returned as `:assets`.  Hence on Windows the code in the `if` block, to create the `ProjectAssetManager` never executes, and Assets can't be opened via the SDK.

Likewise, further down is:
```java
Path assetsPath = gradle.getSubProjects().get("assets").toPath();
```

which also need to take into account the the Assets sub-project will be called `:assets` on a Windows box.

Doing a quick hack to change these 2 to `:assets`, the Asset node was created in the SDK project view, and it was possible to open Assets such as Scenes:

![SceneExplorerWorking](https://user-images.githubusercontent.com/44434629/189528894-c99297e0-024f-42c1-9570-a719f985e1d1.PNG)

peedeeboy:
As this isn't a problem in SDK3.3, I would sumise that there has been a change to the NB Gradle APIs that now needs to be taken into account?

tonihele:
The Gradle template in 3.3 was broken. It was changed for 3.4 and further
in 3.5 (current development). So in all of this versions it is different
structure... or setup.. or both.

tonihele:
Oh and.. The Netbeans you are using... You won't automatically get the new Netbeans, it should be 15 now in master. But you need to delete the `netbeans` folder and `nbproject/private/platform-private.properties`. After next build, a new Netbeans should be at your disposal.

Need to be careful with this if the API indeed changes.

peedeeboy:
@tonihele - your hunch was right, the reason I saw different behaviour between Windows and Linux is because my copy of the SDK repo on Linux was still using a different underlying version of NB.

[...]

PR #380, opened 2022-09-11
#379 - Fix 'File is not part of a jME project'

peedeeboy:
Fixes (hopefully) #379 

Problem was caused by newer underlying version of Netbeans Gradle tooling returning the Asset sub-project name as `:assets` rather than `assets`


tonihele:
Looks like it works. And still works in Ant projects as well as expected. Super! Thank you!

Issue #355, opened 2022-07-24
Upgrade icons to modern SVGs


tonihele:
Upgrade icons to modern SVGs. SVGs are supported ever since Netbeans 11.x. With all the HiDPI stuff and all that, I propose a little bit of a facelift. Take SVGs into use as icons for scaling.  Fortunately we don't have a lot of these.

Icons are here with the constants that need to be changed as well:
https://github.com/jMonkeyEngine/sdk/tree/master/jme3-core/src/com/jme3/gde/core/icons

At least I don't posses the capabilities to create new icons but there should be plenty available in the Internets. Just need to be careful with the licenses.

neph1:
Good idea. I've been using icons from here, lately: https://game-icons.net/tags/gui.html
May not have all needed, but can be combined in inkscape, for example. Liberal license.
If I get the time I might dig into this.

neph1:
More icons here:
jme3-core/src/com/jme3/gde/core/editor/icons
For the shader node editor, it seems.

tonihele:
> I did a naive attempt at replacing a couple of icons, but just loading an svg didn't work. It's either not shown, or extremely small, like a pixel.

Could it be that the format of the SVG is not in the form expected? Needs to have width and height specified it seems.

https://github.com/apache/netbeans/pull/1278/files#diff-9ba09915eb7eaa1837fdd0943744d49bc050900b2a809f09db8fe6c320b81f69R32

neph1:
This is what it looks like:
`<svg style="height: 512px; width: 512px;" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512">`
I also tried various heights/widths and changed the viewBox, as well as scale, but they didn't seem to change anything.

neph1:
Hmm, I had checked that PR, but didn't think of looking for supplied svg's. They do look different. I'll see if I can wrangle the ones I have to fit that format.

neph1:
Bingo, I guess:
![image](https://user-images.githubusercontent.com/7988802/183471557-a5e7a037-99d6-494d-85d0-7e6745ca8d42.png)
I had to be specific about the width/height being 16px for it to fit.

I've been bit conservative and selected icons that look like the existing ones, but it would be a good opportunity for someone with UI inclinations to help pick new/better icons. I'm pretty sure (and glad) we won't find anything looking like the chimp icons.

tonihele:
[...]

PR #385, opened 2022-09-13
Replace icons

neph1:
A first installment of svg icons for core. Some difficult to find/make are still left. Each module then has its own set of icons, usually just a few. I've tried to stick to the theme/look of the old ones, but some liberty has been taken in a few cases.
Would have been nice to keep color and size outside the svg's, but not sure how to do that.

A bunch of empty merge commits refuse to go away. I'll squash this when merging.

#355


tonihele:
Cool cool. The licensing part is ok?

tonihele:
> Cool cool. The licensing part is ok?

Ah on the javadoc. Yeah, looks to satisfy the requirements I suppose.

neph1:
I think we can probably put it somewhere more visible too. There seems to be some credits on the About page already, so maybe there. I don't know if we need to be more specific, though, like iconX by authorY.

tonihele:
> I think we can probably put it somewhere more visible too. There seems to be some credits on the About page already, so maybe there. I don't know if we need to be more specific, though, like iconX by authorY.

Good idea, I don't think we need to be more specific.

neph1:
That should conclude SceneComposer and SceneViewer. I tried to get a more coherent look on the icons and also made them larger, since displays have higher resolution than 640x480 nowadays.
Screenshot from 2022-09-28 20-03-13

neph1:
Unlike the bitmaps, the look and feel actually changes the colors of the svg. I had not anticipated that (but it is in essence a good thing?). I had unfortunately tested the colors on the vanilla look and feel, and not dark metal. The blue and yellow look ok, I think, but the green is a bit too saturated. I better retweak the colors.

neph1:
Hmm, I just updated all the colors to look better with the chosen LAF (sed ftw), but I see now that I did have a dark theme in the first screenshot too. I can't select DarkMonkey in the list. Not sure how to proceed.
On a sidenote, I just checked out DarkLaF in the list. A superior LaF imho. Not "compatible" with the color change, though. Is it only DarkMetal that distorts the colors? Because all the others seem to show them as they were intended.

PR #389, opened 2022-10-02
Material Editor overhaul

neph1:
I have also made a suggestion for changes to the layout of the Material Editor.

image

Basically:
Better use the space on the screen.
Bigger preview.
Remove the automatically expanding tabs, because it's difficult sometimes to see which name was connected with which texture when they were at opposite sides of the screen. Very annoying, and unnecessary. (They can still shrink, though).
Removed some buttons which I think are redundant. Instead, click the color or texture icon to change it. (I used the same for the Shader Node Editor).


neph1:
I've come as far as I can with this, I think. There's a lot of space left in there, but I don't know what to fill it with. Thoughts on this "refresher"? Two columns good for options? @tonihele ? @MeFisto94 ?

tonihele:
Good you asked the forum. Looks ok to me, can't really give you much input :(

PR #391, opened 2022-10-07
Fixes_#336_Default_values_from_MatDef

### LINKED ISSUE #336: Default values from MatDef (at least with Shader Nodes) doesn't propagate to material
When you create a Material from a MatDef it's expected that the default values of the MatDef populate the values in the Material. None of the values are visible in the Material source (but the Material looks correct).

neph1:
Fixes #336 Some noise here due to me removing some warnings and formatting the file. Check the comment for the change that matters.
It actually changes the "source" as well. But I guess that is more in line with WYSIWYG, after all.

### DIFF
--- a/jme3-materialeditor/src/com/jme3/gde/materials/EditableMaterialFile.java
+++ b/jme3-materialeditor/src/com/jme3/gde/materials/EditableMaterialFile.java
@@ -145,11 +145,11 @@ public class EditableMaterialFile {
                     }
                     if (lines != null) {
                         MaterialProperty prop = new MaterialProperty();
-                        String name = lines[0].trim();
-                        prop.setName(name);
-                        if (additionalRenderStates.get(name) != null) {
+                        final String propName = lines[0].trim();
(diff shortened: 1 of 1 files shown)


jme3-materialeditor/src/com/jme3/gde/materials/EditableMaterialFile.java
```diff
@@ -240,18 +240,23 @@ private void checkWithMatDef() {
...
+                                matDefEntries.add(propName);
+                                MaterialProperty prop = materialParameters.get(propName);
                                 if (prop == null) {
                                     prop = new MaterialProperty();
-                                    prop.setName(name);
-                                    prop.setValue("");
+                                    prop.setName(propName);
                                     materialParameters.put(prop.getName(), prop);
                                 }
+                                // property with set value
```
neph1, line 253: This is the functional change of this PR.

tonihele (approved):

PR #393, opened 2022-10-07
Revert "#264 Removal of ogrexml. Unsure of whether I've caught everyt…

neph1:
…hing, but it builds and loads."

This reverts commit 0731b0af25eca9bee7b9138517650124b1c2868f.

Apparently it didn't build nor load. I've looked for obvious fixes, but didn't find any, so here's a revert PR until I can dig into this further.


neph1:
:sweat:

PR #394, opened 2022-10-07
fix ogre xml dependency

neph1:


neph1:
I think this fixes it, at least it loads locally. But I don't have time to check it now. Will do later (unless someone else does it first)

PR #398, opened 2022-10-13
NormalViewState appstate and hook up for SceneViewer

### LINKED ISSUE #200: Add Normals Debug
Much like this: 
https://github.com/jmonkeystore/intellij-integration/blob/master/src/main/java/com/jmonkeystore/ide/jme/scene/NormalViewerState.java

The button should be close to the Wireframe Buttons and we should add a new geometry instead of changing the material as well

neph1:
This is using a temporary icon. I'll fix that once https://github.com/jMonkeyEngine/sdk/pull/385) is merged, rather than building this on top of the other branch.

image

Normal length could be debated. Currently it's 0.2 units.

Fixes #200

### DIFF
--- a/jme3-core/src/com/jme3/gde/core/scene/SceneApplication.java
+++ b/jme3-core/src/com/jme3/gde/core/scene/SceneApplication.java
@@ -36,6 +36,7 @@ import com.jme3.gde.core.assets.AssetData;
 import com.jme3.gde.core.assets.AssetDataObject;
 import com.jme3.gde.core.scene.controller.AbstractCameraController;
 import com.jme3.gde.core.scene.processors.WireProcessor;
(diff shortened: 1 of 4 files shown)


neph1:
Maybe it should have support for moving stuff around. Currently you'd have to toggle it off and on to see the changes.
Maybe also a refactoring of SceneApplication should be considered if we add more functionality, moving more to AppStates.

neph1:
Now it updates the debug transforms every update. Unnecessarily expensive? I considered doing some checks, like debug.transform != geometry.transform, but the checks themselves might be just as expensive.
I'd like to check for if a geometry is added or removed and mirror that in the debug view, but while I thought checking for removed ones would be simple, it didn't work as I had expected.

PR #400, opened 2022-10-18
#324 - changes VarType to JComboBox with hardcoded values for Shader Node wizard

### LINKED ISSUE #324: Create ShaderNodeDefinition - "Type" could be drop down menu.
Currently you need to type the "Type" of inputs and outputs as text. Would be neat to select from a drop down menu instead.

neph1:
Hope I didn't miss anything. I feel like those values should be available somewhere, but don't know where. 
(Adding the default "float" to make the row visible. Very confusing otherwise and you tend to click on "Add" several times)

Fixes #324


MeFisto94:
Hmm, the VarTypes are openGL keywords. But then the parser probably has them as well.
However, I guess it's not too bad we're using a manual list.
One thing though: From the commit (and PR title) it's not clear that this is related to shader node variable type definitions

PR #417, opened 2022-11-02
Include the test data

### LINKED ISSUE #266: Showing HDR Environment is broken
```
com.jme3.asset.AssetNotFoundException: Textures/Sky/Path.hdr (Flipped)
    at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:382)
    at com.jme3.asset.DesktopAssetManager.loadTexture(DesktopAssetManager.java:404)
    at com.jme3.util.SkyFactory.createSky(SkyFactory.java:291)
    at com.jme3.gde.core.scene.SceneApplication$11.run(SceneApplication.java:640)
```

I wonder how this could happen, given that I just added this feature very recently.
The file should stem from testdata.

tonihele:
Resolves #266

### DIFF
--- a/jme3-project-testdata/nbproject/project.xml
+++ b/jme3-project-testdata/nbproject/project.xml
@@ -7,6 +7,10 @@
             <suite-component/>
             <module-dependencies/>
             <public-packages/>
+            <class-path-extension>
+                <runtime-relative-path>ext/jme3-testdata.jar</runtime-relative-path>
+                <binary-origin>release/modules/ext/jme3-testdata.jar</binary-origin>
+            </class-path-extension>
         </data>
     </configuration>
 </project>


tonihele:
Blah, seems that this works one time only. The real solution is to do this in Gradle, like the others....

tonihele:
Doesn't work... Needs investigation...

neph1:
I took a look at this issue using this PR as a base. I didn't get very far, but I did notice that the jme3-project-testdata path doesn't use the more common "release/modules/ext/", but rather "release/libs". Not sure if it means anything, and didn't work even though I changed it. :shrug:

neph1:
Maybe I'm misunderstanding things, but I think it actually works.
image

The lib just needs to be moved like in my comment above (to match the xml)

PR #422, opened 2022-11-05
Bugfix/issue 421

### LINKED ISSUE #421: NPE from Scene Composers Show normals
NPE from Scene Composers Show normals. And I think it is using the older tangent generator.

image

tonihele:
Resolves #421 

It appears I had some curves and all that in scene editor and it just crashed. Better to skip all that doesn't have normals to display.


tonihele:
Also went through some annoying tooltips, everything in capital first letter except these. And the font ones plain wrong. I though that this is so small so I just integrated these to this bugfix. It is wrong of course..

neph1 (approved):
Good catch!

PR #438, opened 2022-12-12
propose to fix #266 by moving installed jar

### LINKED ISSUE #266: Showing HDR Environment is broken

neph1:
This is essentially https://github.com/jMonkeyEngine/sdk/pull/417
I don't know enough about the installation process to know if it actually fixes the problem, but I noticed that if I used https://github.com/jMonkeyEngine/sdk/pull/417 and then moved the lib to the folder it actually points to, it worked.

### DIFF
--- a/jme3-project-testdata/nbproject/project.xml
+++ b/jme3-project-testdata/nbproject/project.xml
@@ -7,6 +7,10 @@
             <suite-component/>
             <module-dependencies/>
             <public-packages/>
+            <class-path-extension>
+                <runtime-relative-path>ext/jme3-testdata.jar</runtime-relative-path>
+                <binary-origin>release/modules/ext/jme3-testdata.jar</binary-origin>
+            </class-path-extension>
         </data>
     </configuration>
 </project>
--- a/jme3-project-testdata/src/com/jme3/gde/project/testdata/jme3-test-data.xml
+++ b/jme3-project-testdata/src/com/jme3/gde/project/testdata/jme3-test-data.xml
@@ -10,7 +10,7 @@ and open the template in the editor.
     <localizing-bundle>com.jme3.gde.project.testdata.Bundle</localizing-bundle>
     <volume>
         <type>classpath</type>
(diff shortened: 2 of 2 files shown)


tonihele:
If it works after you do clean and build. Restart IDE and do it again. Then... it is maybe fine? We can also do a test release at some point to see if it worked or not. Some small fixes already accumulated after last release?

neph1:
It didn't. Needs investigation... I'll keep trying though. Something made it work.

tonihele:
Yeah, for me it was like that too. Works when you do it. But.... somehow not permanently.

neph1:
If it bleeds, we can kill it.

neph1:
I did a little bit of further testing with the release/libs path, to no avail. I don't understand why this works but not the other. But unless there are objections, I'll probably merge this some time during the weekend.

tonihele:
If it works, I'm good with it. I suspect anyway if we get the Maven done. This is obsolete. Maybe.

tonihele:
Oh, did you by the way test the test project? Doesn't it use the test data as well on some of the tests?

neph1:
Which project would that be?

neph1:
Ah, the JmeTests project. No I didn't, and of course it didn't work when I did.

neph1:
fix here: https://github.com/jMonkeyEngine/sdk/pull/440

PR #440, opened 2022-12-17
fix JmeTests project requirement

neph1:
build.gradle might not be required, but i noticed a path that wasn't updated. the fix is in jme-test-data.xml


neph1:
The build fails because it doesn't find nb 15. Did I do that? https://github.com/jMonkeyEngine/sdk/actions/runs/3719995605/jobs/6309204397

tonihele:
When a new release comes, they put the old to some legacy folder. It is a bit annoying... One thing that Maven will solve again! :)

PR #441, opened 2022-12-17
update to netbeans 16 to fix build

neph1:


tonihele:
NB 16 is a bit broken. So better not release with it. Some Gradle stuff i.e. broken and they are working on a hotfix. Did you also check that does the installer need a upgrade? Or maybe doesn't matter if we are not going to release with 16.

tonihele:
https://github.com/apache/netbeans/pulls?q=is%3Apr+milestone%3ANB16u1+is%3Aclosed

Maybe it is released already...

neph1:
Still safer to stay with nb15? I don't know. You seem to know more so you take the decision 😁

tonihele:
https://github.com/jMonkeyEngine/sdk/blob/master/build.gradle#L559 we have this @MeFisto94 magic in-place. I'm in favour of always taking up-to-date NB. Given there is no broken peaces (that we know of). And now if they already released the patch, we will also release with that. So that is ok.

tonihele:
The process of upgrading NB is the following:

- Gradle build, the new download URL
- https://github.com/jMonkeyEngine/sdk/pull/366, the harness, haven't changed recently but should always check (instructions: https://github.com/jMonkeyEngine/sdk/tree/master/harness-override)
- https://github.com/jMonkeyEngine/sdk/pull/361, installer, also doesn't always change, just matter of copying the files from the downloaded NB

We should really document this. Although.... again, with Maven, obsolete :)

neph1:
So something like this: https://github.com/jMonkeyEngine/sdk/pull/442
(harness hadn't changed)

Issue #373, opened 2022-08-26
ProjectAssetManager doesn't load resources from dependency Jar files in Gradle projects


peedeeboy:
As mentioned by @yaRnMcDonuts in #305 :

>Yes I notice that as well. The drop down menu contains all my own shaders located in my projects assets folder, but none of the MatDefs from dependencies such as JME core or Terrain.

>I am also noticing that the material editor begins to work again with my GradleTemplate based project when I open up a material using a MatDef from my assets folder, so I believe if this issue is resolved then it will also resolve the other issue with Gradle projects in the SDK where the visual Material Editor doesn't work for the stock Lighting.j3md and PBRLighting.3md MatDefs

This happens because the Materials Editor does two things:

1) Scans for files in the project directory tree with a `.j3m` extension
2) Defers to the [ProjectAssetsManager](https://github.com/jMonkeyEngine/sdk/blob/master/jme3-core/src/com/jme3/gde/core/assets/ProjectAssetManager.java) to load dependencies from .jar files in the project

Part 1) still works, and will find .j3m files in the `/assets` folder/sub-project, but the `ProjectAssetManager` class is using NB's Java Project API to the files on the classpath, and its not returning the files for Gradle projects.  `ProjectAssetManager` needs to be extended to be "Gradle-aware" and probably use NB's [Gradle Java Projects API](https://bits.netbeans.org/dev/javadoc/org-netbeans-modules-gradle-java/org/netbeans/modules/gradle/java/api/GradleJavaSourceSet.html) to locate the dependency .jars (after a priming build) in Gradle projects and load the materials from them.

This would probably fix a bunch of other things that are squiffy in Gradle projects, like the textures browser etc.

tonihele:
And then there is this, not maybe 100% related but:
WARNING [com.jme3.gde.core.assets.ExternalChangeScanner]: File is not part of a jME project but tries to find original model...

If I create a scene file to the Gradle template... It doesn't seem to understand it is a jME project or something....

peedeeboy:
> If I create a scene file to the Gradle template... It doesn't seem to understand it is a jME project or something....

Yes - I noticed that. Similar, in that I think it is to do with the `ProjectAssetManager`, but I don't think directly related.

With SDK3.4, if I create a new Gradle project, the Asset node gets created in the Project view.  This appears to be when the `ProjectAssetManager` then gets initialised.  Opening Scenes etc. works in this new project.

[...]

PR #448, opened 2023-01-02
Improve ProjectAssetManager's Gradle support

peedeeboy:
Hopefully goes some way to fixing #373 

A couple of bits that could use further refinement:

1. When you first create a new Gradle JME project, NB will prompt you for a priming build - due to changes in NB's Gradle API in latest version, because our Gradle project is created by unziping a file rather than using NB's Gradle API, it is untrusted at this stage and the priming build doesn't run.  This means the user need to run Gradle build again manually before resources will be loaded off the Gradle dependencies - the real fix for this needs to be in the Gradle template.

2. Hard coded to use look at the `main` sourceset - if user has defined custom sourcesets in their `build.gradle` these will get ignored

Could do with some testing, inc. making sure nothing in Ant based projects is now broken...

### DIFF
--- a/jme3-core/src/com/jme3/gde/core/assets/ProjectAssetManager.java
+++ b/jme3-core/src/com/jme3/gde/core/assets/ProjectAssetManager.java
@@ -37,6 +37,7 @@ import com.jme3.asset.AssetManager;
 import com.jme3.asset.DesktopAssetManager;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
+import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -45,9 +46,11 @@ import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.netbeans.api.java.classpath.ClassPath;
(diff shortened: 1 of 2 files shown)


peedeeboy:
Spotted a couple of typos.  Will push a quick update tomorrow

neph1:
Tested:
Loading matdefs in gradle project works great! :+1: 
Sadly, the ant projects seem to have the problem now instead.

jme3-core/src/com/jme3/gde/core/assets/ProjectAssetManager.java
```diff
@@ -478,16 +524,19 @@ private List<String> collectDependenciesFilesWithSuffix(String suffix, List<Stri
...
                 ClassPathItem classPathItem = classPathItemsIter.next();
                 FileObject jarFile = classPathItem.object;
-
-                Enumeration<FileObject> jarEntry = (Enumeration<FileObject>) jarFile.getChildren(true);
-                while (jarEntry.hasMoreElements()) {
-                    FileObject jarEntryAsset = jarEntry.nextElement();
-                    if (jarEntryAsset.getExt().equalsIgnoreCase(suffix)) {
-                        if (!jarEntryAsset.getPath().startsWith("/")) {
-                            list.add(jarEntryAsset.getPath());
+                if(FileUtil.isArchiveFile(jarFile)) {
```
neph1, line 527: It's this change that causes MatDefs not to show in ant projects

peedeeboy: Hye @neph1 ! Thanks for testing :+1: 

This is exactly the bit I thought might break Ant :cry: I can't remember exactly, but from memory I think from the NB Ant Project API we get the dependent jars with `jar://` URLs, but from Gradle Project API we get back `file://` URLs.

Basically, from Ant projects we explicitly knew they were Jar files, whereas in Gradle I'm just treating them here as a regular .zip file and trying to add everything in it for classloading.

I should have a bit of free time this weekend to look at this :+1:

peedeeboy: @neph1 - try now please 😺 

p.s. please squash if/when you merge this is - I tidied up a couple of typos and the fact I forgot to switch the logger back to being less 'chatty'

neph1 (approved):
Tested MatDefs in Ant project: OK
Tested MatDefs in Gradle project: OK

:clap:

PR #458, opened 2023-01-22
Do not overwrite Editor Profile on every load

### LINKED ISSUE #457: Editor color profile stuck to Dark Monkey
Editor color profile stuck to Dark Monkey. This is a regression. I use mostly Flatlaf Light all around. But try setting it now in current master. After restarting the IDE it is back to Dark Monkey...

image

peedeeboy:
Fixes #457 

In order to get FlatLaf Dark to work "first time" without needing an IDE reset to fix the colours of the Option header bar etc. I had to move setting the default LaF from the static initialisation block to the `validate()` lifecycle hook.

Setting the Editor profile doesn't work in the `validate()` block though (too early in IDE startup), so I moved it to the `restored()` lifecycle hook which gets called later - completely missing that it was no longer behind an `if` so Editor profile was getting set to Dark Monkey colours every time the IDE started.

This fixes that by setting a static boolean so that we only set the Editor profile if we are also setting the default LaF (i.e. the first time we load the darkmonkey module).

### DIFF
--- a/jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/Installer.java
+++ b/jme3-dark-laf/src/org/jme3/netbeans/plaf/darkmonkey/Installer.java
@@ -12,6 +12,8 @@ import org.openide.util.NbPreferences;
 
 public class Installer extends ModuleInstall {
 
+    private static boolean isFirstInstallation = false;
+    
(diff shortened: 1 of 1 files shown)


neph1 (approved):
Tested :+1:

PR #462, opened 2023-02-01
Fix NPE and reduce nesting

tonihele:
Not sure if the Ant projects should just continue to the few lines after the seemingly Gradle specific thing... But Ant project assets seem to work ok (shows the yellow icon and can open to scene explorer). Having Ant projects open, this resulted in nasty NPE.


neph1:
I didn't notice the issue when testing, but perhaps it was because I had cleaned and only opened the relevant projects, ie "Happy flow"

tonihele:
It doesn't look the error actually causes any ill effect. It is just present on the logs. Noticed randomly while testing the Maven version thing.

tonihele:
Ok, I checked the history, so yeah, now the NPE is gone an otherwise everything is as before for other than Gradle types. I think it is correct now.

PR #464, opened 2023-02-04
Feature/issue 371

### LINKED ISSUE #371: Try to get version information to Gradle template wizard from Maven
Try to get version information to Gradle template wizard from Maven. Now that we have the wizard, it is awesome. But my biggest fear is that it will become a chore. So I would like it to fetch the version information from Maven central automatically. The release notes might be non-existent or lacking with this. But at least it would live on without SDK releases.

Netbeans at least have some MavenUtils of sort available. I wonder if they can be directly used to pull something like this off...

tonihele:
Resolves #371 

Alright.. Now, if Maven API is up, we initialize the user project with latest versions of libraries. And automatically display latest versions of jME to choose from. If Maven API is down, it doesn't matter, we have a set of hard coded versions to choose from. This is just an enhancement that is provided for our and users' convenience.

Maven API wasn't working that great this past two weeks. Now it seems to be up again. How this is designed, is that there is **zero wait for the user**. So there might be inconsistent results in users' perspective as we might catch latest versions at some random point in time.

In my experience, the Maven API either responds immediately or it is simply down. Meaning that this strategy is viable. The Maven is accessed in highly parallel way, zero locks still. And only once per lifetime of the SDK instance.

Ok, the caveats:

- User sees some warnings for not able to get latest versions of some libraries (due to API error or as guaranteed since not all the libraries while writing this are accessible from Maven Central)
- Even for jME libraries we can't guarantee a working combination on all versions. Libraries such as `test-data` and `awt-dialogs` might be missing from some versions, yet they are hard coded to be included (not the latter, yet)

edit: added library version filtering to avoid weird version types (minie and heart)


tonihele:
> * The latest versions might be in some cases little bit off.. Or depends on the use case.. i.e. `com.github.stephengold:Heart:8.3.0+for36` and `com.github.stephengold:Minie:7.2.0+mt` (we can maybe ask Stephen to consider adjusting his build scripts so that the recommended package is uploaded last...)

Ok, this is kinda ridiculous demand. We can fix this to be better...

tonihele:
I'll merge this in 12h if no objections. I think it is production ready.

I still tweaked the version handling to more straightforward model in code.

PR #467, opened 2023-02-06
Fix #463 by allowing other texture settings and only changing Flip an…

### LINKED ISSUE #463: Material parsing problem
As reported in: https://hub.jmonkeyengine.org/t/textures-materials-and-sdk/46413/7

Some material properties have problem parsing:
Error while locating rapRepeat_S WrapRepeat_T MinBilinearNoMipMaps "Models/Proto/body_diffuse.jpg
com.jme3.asset.AssetNotFoundException: rapRepeat_S WrapRepeat_T MinBilinearNoMipMaps "Models/Proto/body_diffuse.jpg (Flipped)
	
I did some changes to make the editor parse the values of the materials correctly a while back. It could be related to that. Or, the SDK has had issues parsing the texture line before, the quotation marks, several years ago. Maybe it's a left over from that.

neph1:
…d Repeat

This is a preliminary fix. Not ready to merge yet.
I'd like a second set of eyes before merging.

So the cause was that all those "new" settings have never been implemented in the Material Editor.
Implementing all those is an undertaking I'm not ready to do right now. So my proposed interim solution is to not care about other settings than "Flip" and "Repeat", but not overwrite them either. This allows advanced users to keep modifying their materials in the Material Editor (but preview might be wrong? I'm not familiar with the settings).

Edit: Still need to update TexturePanelSquare. Only used in shader editor, I think

### DIFF
--- a/jme3-materialeditor/src/com/jme3/gde/materials/multiview/widgets/TexturePanel.java
+++ b/jme3-materialeditor/src/com/jme3/gde/materials/multiview/widgets/TexturePanel.java
@@ -54,7 +54,8 @@ public class TexturePanel extends MaterialPropertyWidget {
                         if (texPreview == null) {
                             texPreview = new TexturePreview(manager);
                         }
-                        texPreview.requestPreview(stripQuotes(textureName), "", 80, 25, texturePreview, null);
+                        final String[] textureNameComponents = textureName.split(" ");
+                        texPreview.requestPreview(stripQuotes(textureNameComponents[textureNameComponents.length-1]), "", 80, 25, texturePreview, null);
                     } catch (AssetNotFoundException a) {
                         Logger.getLogger(MaterialEditorTopComponent.class.getName()).log(Level.WARNING, "Could not load texture {0}", textureName);
                     }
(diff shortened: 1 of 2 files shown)


jme3-materialeditor/src/com/jme3/gde/materials/multiview/widgets/TexturePanel.java
```diff
@@ -54,7 +54,8 @@ public void run() {
                         if (texPreview == null) {
                             texPreview = new TexturePreview(manager);
                         }
-                        texPreview.requestPreview(stripQuotes(textureName), "", 80, 25, texturePreview, null);
+                        final String[] textureNameComponents = textureName.split(" ");
+                        texPreview.requestPreview(stripQuotes(textureNameComponents[textureNameComponents.length-1]), "", 80, 25, texturePreview, null);
```
neph1, line 58: Split and take last segment to get actual texture name

jme3-materialeditor/src/com/jme3/gde/materials/multiview/widgets/TexturePanel.java
```diff
@@ -68,19 +69,21 @@ private String stripQuotes(String s) {
...
+            propertyValue = propertyValue.replaceFirst("Flip ", "");
+        }
+        if(repeat && !propertyValue.contains("Repeat ")) {
+            propertyValue += "Repeat ";
+        } else if(!repeat) {
+            propertyValue = propertyValue.replaceFirst("Repeat ", "");
         }
+        propertyValue += textureName;
+        property.setValue(propertyValue);
+        texturePreview.setToolTipText(propertyValue);
```
neph1, line 86: Don't replace whole line, just modify Flip and Repeat, if they exist

jme3-materialeditor/src/com/jme3/gde/materials/multiview/widgets/TexturePanel.java
```diff
@@ -281,19 +284,16 @@ protected void readProperty() {
...
+                    textureName = textureName.replaceFirst("Flip ", "").trim();
+                } 
+                if (textureName.contains("Repeat ")) {
                     repeat = true;
-                    textureName = property.getValue().replaceFirst("Repeat ", "").trim();
-                } else {
-                    textureName = property.getValue();
+                    textureName = textureName.replaceFirst("Repeat ", "").trim();
                 }
+                property.setValue(textureName);
```
neph1, line 296: Simpler?

neph1:
The following tests have been performed:

- Loading/saving settings on the texture line
- Editing Flip/Repeat with/without other settings on the texture line
- Loading material in application with/without various settings

If someone wants to test this themselves, please do. Otherwise I'll merge this after an unspecified amount of time.

PR #473, opened 2023-02-19
Feature/issue 471

### LINKED ISSUE #471: Update Java to 17
We could try updating Java to 17. Bundled Java as well as the build.

While this is just mainly maintenance. I have a small hope that it would fix the slow file open dialog I've been hearing about. Though looking at this issue we have tried to report it year ago. And looks like newer Java 8 broke it, Java 11 didn't fix it then... how about Java 17?
https://issues.apache.org/jira/browse/NETBEANS-1537

tonihele:
Resolves #471 

Hmm, yeah. This can be run locally now. Probably releasing needs a bit tweaking. Pack200 doesn't exist. These params added to project.properties need to go to the configuration of the release.... But I don't know how really to test this.


tonihele:
It does work, I tested it. On Linux though, built the installers and it works. There is couple of things I learned in the process and I wish to share them with you.

`nbproject/project.properties` file has a property `run.args.extra`. Basically JVM args when you run the SDK. These used to get copied to the actual installation (`etc/jmonkeyplatform.conf`) back in Netbeans 8 age. I'm not exactly sure when this broke down. But it did broke with newer NB releases as they cleaned the default harness app.conf file not to have any memory params or anything like that.

With this PR this functionality is restored. With that I stripped the old JVM args that should not be used anymore. And replaced those with some that are needed with modern JDKs. At least for awhile now before Netbeans fixes their code to be more native to modern SDKs.

This however solves this issue I always wondered. The issue of graphic corruption (https://hub.jmonkeyengine.org/t/solved-sdk-3-3-and-gui-issue/45642/24). I'm also a victim of this. But it hasn't happened now again, so I did not include the flags. This bug never happened to me when debugging or running the SDK via Netbeans. And I finally found the answer why, most likely it is this. Old SDK versions worked because the args were copied. I did not however now include the parameters. Lets track this. It is maybe even fixed and no need for those parameters anymore.

The flags needed by modern JDKs are all in https://github.com/apache/netbeans/tree/master/nbbuild/jms-config. But I only included what I needed. Or rather what I encountered. This is the suggested way anyhow.

tonihele:
I'll merge this in 48h if no objections.

PR #475, opened 2023-03-05
Netbeans 17

### LINKED ISSUE #474: Upgrade to Netbeans 17
Upgrade to Netbeans 17. 16 wasn't apparently that great. At least 17 as of yet doesn't have known critical updates in the making. Anyway, latest and greatest for the release. We aren't holding back.

tonihele:
Looks to be working. Resolves #474


tonihele:
I'll self integrate this in 24h if no objections arise.

PR #476, opened 2023-03-10
Update to jme 3.6

neph1:
I didn't commit the psd's since it would mean the layers would become rasterized. Not sure if we need them going forward.
This can probably be merged pretty soon, and then I'll update to stable once that is available.


tonihele:
Looks good to me!

PR #481, opened 2023-03-22
update to 3.6.0-stable and ignore META-INF entry when making project.xml

neph1:
Not sure if this is a great solution, but it seems to work when i build locally.


neph1:
@tonihele @MeFisto94 Could either of you check if you find this to be OK? I want a second opinion before merging :)

tonihele:
At least locally on Linux it works fine! And I would say finally because the LWJGL 2 previously blocked me on working on Linux. I simply tested ./gradlew run on this branch.

We can revert it also later if some problems arise. Maybe just better to go forward. Have a jME 3.6 release with working Gradle assets and all that goodness. Hoping Java 17 would solve the file dialog slowness issues...

Issue #176, opened 2018-11-07
Attaching a light probe to a geometry throws exception and cant be removed from node.


mitm001:
Attaching a light probe to the geometry will throw this exception.

SEVERE [com.jme3.gde.core.scene.SceneApplication]: Scene graph is not properly updated for rendering.
State was changed after rootNode.updateGeometricState() call. 
Make sure you do not modify the scene from another thread!
Problem spatial name: PBRCube
java.lang.IllegalStateException: Scene graph is not properly updated for rendering.
State was changed after rootNode.updateGeometricState() call. 
Make sure you do not modify the scene from another thread!

Understandable, but the attached lightprobe cannot be removed without restarting the SDK after delete.

grizeldi:
Reproduced. This leaves SceneComposer completely unusable for any file (not just the one where the mistake was made) until the sdk is restarted.

PR #490, opened 2023-04-04
Work around for #176 "Scene graph is not properly updated for renderi…

neph1:
…ng" when adding LightProbe to a Geometry.

It seems the issue happens inside LightProbeFactory. I think this is an acceptable solution for now, since the consequences of the error are pretty bad.

If someone really wants a LightProbe for just one Geometry, they can replicate it by adding the Geometry to a Node and add the LightProbe to that.

### DIFF
--- a/jme3-core/src/com/jme3/gde/core/sceneexplorer/nodes/actions/NewLightPopup.java
+++ b/jme3-core/src/com/jme3/gde/core/sceneexplorer/nodes/actions/NewLightPopup.java
@@ -48,6 +48,7 @@ import com.jme3.light.PointLight;
 import com.jme3.light.SpotLight;
 import com.jme3.math.ColorRGBA;
 import com.jme3.math.Vector3f;
+import com.jme3.scene.Node;
 import com.jme3.scene.Spatial;
 import java.awt.event.ActionEvent;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


tonihele:
Hmm, or is it that now the modification happens from the Swing thread. And
instead it should be added in jME render thread? With the enqueue.

neph1:
My first test was to comment out:
```
node.addLight(lightProbe);
node.updateModelBound();
```
And it still happened

neph1:
This is probably the culprit:
EnvironmentCamera:
```
@Override
    public void render(final RenderManager renderManager) {
        if (isBusy()) {
            final SnapshotJob job = jobs.get(0);

            for (int i = 0; i < 6; i++) {
                viewports[i].clearScenes();
                viewports[i].attachScene(job.scene);
                renderManager.renderViewPort(viewports[i], 0.16f);
                buffers[i] = BufferUtils.createByteBuffer(
                        size * size * imageFormat.getBitsPerPixel() / 8);
                renderManager.getRenderer().readFrameBufferWithFormat(
                        framebuffers[i], buffers[i], imageFormat);
                images[i] = new Image(imageFormat, size, size, buffers[i],
                        ColorSpace.Linear);
                MipMapGenerator.generateMipMaps(images[i]);
            }

            final TextureCubeMap map = EnvMapUtils.makeCubeMap(images[0],
                    images[1], images[2], images[3], images[4], images[5],
                    imageFormat);
            debugEnv = map;
            job.callback.done(map);
            map.getImage().dispose();
            jobs.remove(0);
        }
    }
```

I guess proper handling would be to let the LightProbe finish its rendering and then attach it. Or just add it and let the user "update the maps" themselves?

neph1:
Acceptable temp fix? I'll merge this soon unless there's further discussion.

Issue #491, opened 2023-04-06
Installer fails on Windows (x64)


neph1:
https://hub.jmonkeyengine.org/t/error-installing-jmonkeysdk/46659/9

I’m trying to install the new version of the SDK Release 3.6 [pre] sdk and I get the following error:

    Cannot prepare bundled JVM to run the installer

my device:
11th Gen Intel(R) Core™ i5-1135G7 @ 2.40GHz 2.42 GHz
64bit processor
windows 11 :frowning:

(Reproduced. No logs)

tsatatwer:
I have the same exact problem windows 10  jdk 17 64bit

tonihele:
I created a bug ticket for Netbeans itself https://github.com/apache/netbeans/issues/5783. The installer is here: https://github.com/apache/netbeans-native-installers. If it helps to decipher what is going on.

tsatatwer:
There is a work around (BUT missing JME libraries)  on windows 10:
open power shell 
change directory where the installer  **jmonkeyplatform-windows-x64.exe**
**jmonkeyplatform-windows-x64.exe  --extract** 
**java -jar  bundle.jar**
It will install **BUT missing JME libraries**

MeFisto94:
Another workaround would be to use the platform independent zip and run it's jar like that (`java -jar jMonkeyPlatform.jar` or something)

tsatatwer:
@MeFisto94  where is  **jMonkeyPlatform.jar** ??? witch path??

tonihele:
We now insert a fake `unpack200.exe` to the bundled Windows JDKs. This is to satisfy Netbeans Installer (NBI) check for a valid JDK. Also all pack200 should now be disabled from the project.

MeFisto94:
Sorry, thought we have a jar, but still: 
```sh
wget https://github.com/jMonkeyEngine/sdk/releases/download/v3.6-stable-sdk3/jmonkeyplatform.zip
unzip jmonkeyplatform
./jmonkeyplatform/bin/jmonkeyplatform (or the .exe variants)
```
can always be used as a workaround.

PR #492, opened 2023-04-09
Fix Windows installer

tonihele:
pack200 was removed with Java 14. We are on Java 17. No-go. I did not test this but looks like the Windows installer falls to using the bundled JDK (17) to try unpack pack200 and fails. I have no idea why it works on Linux.


tonihele:
Resolves #491 

Yeah, I tested this. NB Installer (NBI) seems to check for unpack200.exe from the bundled JDK. It checks this before it even looks that is there anything to unpack. We don't, at least not anymore. All the pack200 are now turned off. The fake unpacker needs to be bundled until NBI is fixed.

PR #493, opened 2023-04-09
Bump Android build to Java 8

tonihele:
Hmm yeah, no one is going to have Java 7 installed. Set to Java 8. I understand this is the version you want to use for Android. This just hopefully bumps the out-of-the-box compatibility forward. Although.... I'm not sure at all. I am totally out of the water here...


tonihele:
Or is it:
https://developer.android.com/studio/write/java8-support

PR #501, opened 2023-04-19
fixes texture name parsing issue in texturepanel

### LINKED ISSUE #500: Material editor name parsing fails on spaces
https://hub.jmonkeyengine.org/t/material-ust-basecolor-not-found-by-asset-manager/46709

neph1:
don't split on " ". Now splits on "\"" instead to get the actual name. Makes the stripQuotes method obsolete Old texture property wasn't cleared when adding a new texture. This looked ok in the editor, but the source file had all the old names as well.

Fixes #500


neph1:
Merge soonish.

PR #506, opened 2023-05-15
Assetbrowser

neph1:
This is a mammoth PR, but there's no other way to do it..


neph1:
I need a break from this now...

Some outstanding issues:

- Gradle project returns a path to 'assets' rather than the project folder.
- Opening a model from the asset browser is much much slower than doing it from the assets node.
- Some space utilization issues in the asset browser panel.
- Previews not updating after being created?

neph1:
@peedeeboy Before I dig into this myself, do you know why this happens?

When I do a 
`projectAssetManager.getProject().getProjectDirectory();`
on an Ant project, I receive the expected project folder. but when I do the same call on a Gradle project, it returns the 'assets' folder?

Pics or it didn't happen!
Ant:

Screenshot from 2023-05-27 08-47-02

Gradle:
Screenshot from 2023-05-27 08-49-50

peedeeboy:
I came across [this comment](https://github.com/jMonkeyEngine/sdk/blob/master/jme3-core/src/com/jme3/gde/core/assets/AssetsLookupProvider.java) from @MeFisto94 when I was looking into the Project Asset Manager previously:
```java
                /* Note: The ProjectAssetManager needs the Subproject's Project,
                 * because otherwise reading assets won't work (as assets is
                 * not in the scope of the parent project). But it still needs 
                 * to be added to the Parent Project's Lookup
                */
                
                /* Complicated code to find the relative path, because the 
                 * assets subproject could be stored in a folder called "foo"
                 * or "bar" and we still need to locate it correctly
                */
```

neph1:
Thanks, then I know it's not a bug. I'll add some 'complicated code' of my own to circumvent it :)

peedeeboy:
I came across this as I when I looked into get the PAM to load from assets from Gradle files on the classpath I was thinking / hoping to refactor ProjectAssetManager out to an Interface / Abstract class and have AntProjectAssetManager and GradleProjectAssetManager, but that turned into a rabbit hole..... 

Probably why nobody else had already done it :(

neph1:
I won't attempt to change anything there, just handle it on the behalf of the asset browser

neph1:
Note to self: Still need to do the ant/gradle assets switch.

PR #538, opened 2023-09-30
update version numbers for libraries in com.github.stephengold

stephengold:
Here are the current recommended versions of my 3rd-party libraries.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,7 +44,7 @@ dependencies {
     corelibs dep("org.jmonkeyengine:jme3-lwjgl:$jmeVersion-$jmeVersionTag", true, true)
     corelibs dep("org.jmonkeyengine:jme3-effects:$jmeVersion-$jmeVersionTag", true, true)
     corelibs dep("org.jmonkeyengine:jme3-blender:3.3.2-stable", false, false) // Pin Pointed until jme3-blender has a dedicated release or support is phased out.
-    optlibs dep("com.github.stephengold:Minie:7.1.0", false, false) // replacement for bullet-native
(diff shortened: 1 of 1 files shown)


tonihele:
Also about your libs.... You had tons of suffixes in some of the libraries, now we are trying to default to one without a suffix. Is that correct?

And we also always take the newest one. The SDK automatically checks for the latest release. If Maven Central API is online :)

stephengold:
I can't think which of my libraries have suffixes, other than Minie.

The version of Minie without suffixes is probably the best all-around choice. It runs on all platforms except iOS and has decent performance.

For troubleshooting, I sometimes suggest using a library of the "debug" type, such as "Minie-7.7.0+debug". The downside of debug-type libraries is they're significantly larger and slower. If the SDK could include the debug-type library *in addition to* plain old Minie, that might have some benefit.

tonihele (approved):

stephengold:
> I can't think which of my libraries have suffixes, other than Minie.

I remembered another example. Sometimes when a new Engine release is "in the works" I publish libraries with suffixes like "+for34" or "+for35" to indicate which version of the Engine is targeted. In those situations, I believe it's obvious which version should be used.

tonihele:
> > I can't think which of my libraries have suffixes, other than Minie.
> 
> I remembered another example. Sometimes when a new Engine release is "in the works" I publish libraries with suffixes like "+for34" or "+for35" to indicate which version of the Engine is targeted. In those situations, I believe it's obvious which version should be used.

We don't have such mechanisms in place that we could automatically make this choice for the user. Currently the code is very simple. Well, good to know anyway if we need to tweak our project creation logic.

PR #539, opened 2023-09-30
add 5 optional libraries from com.simsilica

stephengold:
To help steer users toward Paul Speed's libraries.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,6 +53,12 @@ dependencies {
     corelibs dep("org.jmonkeyengine:jme3-plugins:$jmeVersion-$jmeVersionTag", true, true)
     corelibs dep("org.jmonkeyengine:jme3-terrain:$jmeVersion-$jmeVersionTag", true, true)
 
+    optlibs dep("com.simsilica:lemur:1.16.0", true, true)
+    optlibs dep("com.simsilica:lemur-props:1.2.0", true, true)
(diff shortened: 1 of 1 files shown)


tonihele:
I think these will only come to the Ant projects as selectable. Is that the goal?

stephengold:
The goal is to make these libraries as accessible as possible to SDK users. Toward this end, if there's something we could do for Gradle and Maven users, that'd be great.

tonihele:
Did you try the wizard:
image

stephengold:
not recently

I thought this build.gradle was how 3rd-party libraries (like Wes) were bundled into the SDK for inclusion in user projects. If I was mistaken, then please close this PR without integrating.

tonihele:
No, it is probably fine. Im just trying to make sure what you want to do gets done. This way they go to Ant projects only, then the user has to be smart enough to figure out how to add them to their project.

The Gradle is recommended format. And there is the Gradle wizard which lets the user select all the libraries he/she/it/they/them/apache helicopter is interested in. The selection of packages is here: https://github.com/jMonkeyEngine/sdk/blob/master/jme3-templates/src/com/jme3/gde/templates/gradledesktop/options/AdditionalLibrary.java

stephengold:
> The selection of packages is here

Good to know. Thanks!

tonihele (approved):

PR #545, opened 2023-11-09
Feature/gradle template improvements

tonihele:
- Update Gradle wrapper to 8.4 (to support Java 21 LTS)
- Use application plugin for setting main class and all that
- Make an effort to support Macs and LWJGL 3
- Remove deprecated Gradle stuff


tonihele:
Should be good to go. If nothing else arises I'll self integrate this in 48h.

neph1 (approved):
I didn't look into the mysterious zip-file, but cool! :rocket:

tonihele:
> I didn't look into the mysterious zip-file, but cool! 🚀

That should be the Gradle wrapper upgraded only. The actual build files and all that comes nowadays from the Freemarker template.

PR #566, opened 2024-02-25
Update to Gradle 8.6

tonihele:
Upgrade to Gradle 8.6. Looks like the basics at least work. We need this to try Java 21.


neph1 (approved):
I've had some issues with gradle 8 in regards to some android projects. But if it works, it works :)

tonihele:
> I've had some issues with gradle 8 in regards to some android projects. But if it works, it works :)

This is only for the SDK build, which also uses Ant in the end. Gradle just kinda primes it. The Gradle projects created inside the SDK are already on Gradle 8.

The only change seemed to be needed was not to track the build folder. As Ant manages it, it might not exist at time. And we would get error when it doesn't exist. Doesn't seem to affect anything. These are like optimizations to Gradle that it can use incremental builds (at all? or more efficiently?).

PR #572, opened 2024-03-02
Java 21

### LINKED ISSUE #557: Use JDK 21
Use JDK 21. It should be all possible to replace Java 17 with Java 21. We just need to be careful with the stuff actually now removed from Java, such as Thread.stop() (NavMeshGenerator.java).

NavMeshGenerator should maybe be rewritten to use NB provided thread pools anyway. Unfortunately whatever we do, we can't fix it completely. As changing stop to interrupt doesn't in this case do anything. The library doesn't support it as far as I can remember. Let's just say that is the library's problem.

tonihele:
Resolves #557 

Looks to work ok. I'll do a code cleanup in another branch/PR to not mess this one. Code cleanup to remove some deprecated stuff and why not some simple stuff from basic code analysis.


neph1 (approved):

tonihele:
Future dudes, this was missed https://github.com/jMonkeyEngine/sdk/pull/617.

Issue #562, opened 2024-01-05
If you don't remove a texture, and cancel when selecting one, you will get an error in the SceneComposer


neph1:
The material editor replaces the texture with an empty string, but keeps the field in the file. The SceneComposer obviously doesn't handle this and throws an error.
Remove field completely if no texture selected, possibly when saving.

neph1:
I think this is related to #561

PR #574, opened 2024-03-02
Issue 561

### LINKED ISSUE #561: Editing values in the "Options" tab will make textures lose "Repeat"
Maybe other things.

neph1:
This is a bit of a weird one.
I think it's a regression from when I fixed something related to this a while back. The issue happens when a different texture was changed because the loaded flip and repeat weren't persisted.
I was tempted to rewrite a bit how it functioned, but also wanted to consider extra texture options (iirc that is another issue further down the list).
This refactor handles that, but I'm not sure the rest of the SDK does. At least if you try to enter something manually after the texture name, there will be a warning.
Not sure whether to keep it, or not allow anything after the texture name.
Fixes #561 and I think also #562


tonihele (approved):

neph1:
Cheers :tada: . I'll keep this open since I want to do some more testing before mergin

Issue #568, opened 2024-02-26
NPE when the VehicleWheel node is selected + broken SDK 


capdevon:
NPE when the VehicleWheel node is selected in the scene element view tree. If you try to close and re-open the j3o model afterwards, the model view will always be completely black and you will have to restart the SDK.

![image](https://github.com/jMonkeyEngine/sdk/assets/32201701/186d83e4-c2ab-4722-850d-a4fd0f35937e)

Stacktrace:
```java
java.lang.NullPointerException: Cannot invoke "org.openide.nodes.Node$Property.getName()" because "p" is null
	at org.openide.nodes.Sheet$Set.put(Sheet.java:294)
	at com.jme3.gde.core.sceneexplorer.nodes.JmeVehicleWheel.createSheet(JmeVehicleWheel.java:98)
	at org.openide.nodes.AbstractNode.getSheet(AbstractNode.java:416)
	at org.openide.nodes.AbstractNode.getPropertySets(AbstractNode.java:432)
	at com.jme3.gde.core.sceneexplorer.nodes.AbstractSceneExplorerNode.syncSceneData(AbstractSceneExplorerNode.java:280)
	at com.jme3.gde.core.scene.NodeSyncAppState.update(NodeSyncAppState.java:77)
	at com.jme3.app.state.AppStateManager.update(AppStateManager.java:371)
	at com.jme3.gde.core.scene.SceneApplication.update(SceneApplication.java:309)
	at com.jme3.system.awt.AwtPanelsContext.updateInThread(AwtPanelsContext.java:237)
	at com.jme3.system.awt.AwtPanelsContext.access$200(AwtPanelsContext.java:46)
	at com.jme3.system.awt.AwtPanelsContext$AwtPanelsListener.update(AwtPanelsContext.java:80)
	at com.jme3.system.lwjgl.LwjglOffscreenBuffer.runLoop(LwjglOffscreenBuffer.java:126)
	at com.jme3.system.lwjgl.LwjglOffscreenBuffer.run(LwjglOffscreenBuffer.java:160)
	at java.base/java.lang.Thread.run(Thread.java:1589)
java.lang.NullPointerException
```

neph1:
The class is performing some reflection on the VehicleWheel class. A particular method for getAxle exists, but it takes a Vector3f. I assume there used to be one that didn't, but somehow Minie works differently.
Is it safe to just remove this (getAxle), or should it be handled differently?

The code in question:
[...]

PR #588, opened 2024-03-24
Update CharacterControl to be compatible with Minie

### LINKED ISSUE #587: NPE when adding CharacterControl to the Spatial
Hi guys,
After adding the CharacterControl to the Spatial, the SDK raises an Exception and you can no longer open other models and have to restart the editor.

NPE add CharacterControl

NPE add CharacterControl SDK crash

tonihele:
Had to lose couple of fields. In Minie, these fields have a getter that takes in a parameter. This is not really supported at all.

Resolves #587


tonihele:
I made issue about this https://github.com/stephengold/Minie/issues/39. To my knowledge there is no way to support this kind of design where getter takes a parameter.

neph1:
Related: https://github.com/jMonkeyEngine/sdk/issues/568

neph1 (approved):

PR #592, opened 2024-03-29
Update gradle.yml

tonihele:
Not to repeat allowing PRs break the build... CI build on PRs

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -18,6 +18,7 @@ jobs:
       uses: actions/setup-java@v3
       with:
         java-version: 21
+        distribution: 'temurin'
 
     - name: Validate the Gradle wrapper
       uses: gradle/wrapper-validation-action@v2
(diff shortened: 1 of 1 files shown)


tonihele:
This is a really simple change. I just divided the existing job to two phases. The actual NBM deployment is skipped on PRs. But should be still executed on PUSHes to master as it used to (not tested). PR just builds the SDK. No fancy workflows or matrices where used.

MeFisto94:
Historically we had a separate build that was basically running via travis, that's why GHA only did the deployment build. I think there may have been slight nuanced differences.

I just wanted to check it but apparently we even removed the travis file already

tonihele:
> Historically we had a separate build that was basically running via travis, that's why GHA only did the deployment build. I think there may have been slight nuanced differences.
> 
> I just wanted to check it but apparently we even removed the travis file already

Here:
https://github.com/jMonkeyEngine/sdk/blob/56ce13e34e81c7cdd1b6f224d69a744e5377aa30/.travis.yml

PR #593, opened 2024-04-09
Fix incorrect assignment

TheLivan:
Hi. I simple corrected incorrect assignment variable.


tonihele:
@neph1 I assigned you. Java wise this looks logical but since you know asset browser :)

neph1:
It's assetpacks, not asset browser :)
But still. I don't see how this could be wrong.

PR #596, opened 2024-04-20
Always get the latest JDK version

### LINKED ISSUE #595: Use Adoptium API to always get the latest JDK
Use Adoptium API to always get the latest JDK. We always want the latest JDK of specified major version. I at least don't see why we would hold back on this. I have always done it manually just before releases. We could automate this by using Adoptium API.

See:
https://api.adoptium.net/q/swagger-ui/#/Installer/getInstaller

tonihele:
Resolves #595 

Basically now all releases will get the freshest available JDK. We no longer have the exact version info in the script but I don't think that matters. Locally I just wipe out everything before running the script. And the release is always created in a new empty environment. So the exact version in paths doesn't matter.

### DIFF
--- a/jdks/download-jdks.sh
+++ b/jdks/download-jdks.sh
@@ -11,17 +11,17 @@
 set -e # Quit on Error
 
 jdk_major_version="21"
-jdk_version="0.2"
-jdk_build_version="13"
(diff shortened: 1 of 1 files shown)


neph1:
Nice!

PR #605, opened 2024-05-19
Feature/3.7 branding

tonihele:
Change graphics for upcoming version. I tried Gimp 3 alpha. As it has non destructive layer filters. Akin to the Photoshop color overlay thing. I couldn't get it to work to mimic this. However now the Gimp version of the image is exportable straight. I just hid the original colored Jaime layer and did a colored version of it.

Also the Gimp version didn't have the bottom shape (in Photoshop terms), it is now made straight with painting tools in Gimp.

I still used Gimp 3 to edit the XCF so it might require a newer Gimp to further edit/open. Gimp 3 should be out shortly anyway, I don't see this as a problem.


tonihele:
I'll eventually squash & merge this then if I remember...

PR #610, opened 2024-06-15
Add some opens flags (to allow viewing javadoc)

### LINKED ISSUE #608: Add module exception to allow viewing Javadoc in the SDK
Add module exception to allow viewing Javadoc in the SDK. Seems that the following is needed:
`-J--add-opens=java.base/java.io=ALL-UNNAMED`
`-J--add-opens=java.base/java.lang=ALL-UNNAMED`

See:
https://hub.jmonkeyengine.org/t/ide-accessing-documentation-is-not-working/47704

tonihele:
Didn't actually test this one but there should be no harm in it anyway...

Resolves #608


tonihele:
Ok, the build fails because Netbeans 21 is no longer available... This is highly annoying. We should somehow figure out a permalink to the NB releases. Or switch to Maven :)

PR #612, opened 2024-08-22
Create AnimLayers in the SceneComposer and play animations

neph1:
One little annoying thing... They can't be saved!

https://youtu.be/-hgwB6x-11s 

I really tried getting AnimClips and AnimLayers into two different sub nodes, but it didn't work out.


neph1:
Requires: https://github.com/jMonkeyEngine/jmonkeyengine/pull/2307

PR #623, opened 2024-11-25
fix AppStateExplorer not finding AppStates extending BaseAppState

neph1:
Ref: https://hub.jmonkeyengine.org/t/help-needed-with-j3o-sdk-functions/48036


tonihele:
`//TODO: check with proper casting check.. gotta get TypeMirror of Control interface..`

Yeah, probably good idea. I looked at my AppStates and all of them extend my own wrappers. Many layers before we get to jME classes. But if that is not possible, this is anyway good

PR #624, opened 2024-11-26
speculative fix for failed spatial updates in scenecomposer window

neph1:
Since there's a big:

> //TODO: refresh does not work

and I seem to remember seeing it when doing the ExternalChangeScanner changes, I refactored the methods there to do the same thing in this case.
I haven't actually verified that it works in this instance, yet.

### DIFF
--- a/jme3-core/src/com/jme3/gde/core/sceneexplorer/Bundle.properties
+++ b/jme3-core/src/com/jme3/gde/core/sceneexplorer/Bundle.properties
@@ -1,4 +1,4 @@
 CTL_SceneExplorerAction=SceneExplorer
 CTL_SceneExplorerTopComponent=SceneExplorer
 HINT_SceneExplorerTopComponent=The SceneExplorer provides an Overview over the SceneGraph of your Scene.
-SceneExplorerTopComponent.jButton1.text=Update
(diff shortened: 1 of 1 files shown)


jme3-core/src/com/jme3/gde/core/sceneexplorer/Bundle.properties
```diff
@@ -1,4 +1,4 @@
 CTL_SceneExplorerAction=SceneExplorer
 CTL_SceneExplorerTopComponent=SceneExplorer
 HINT_SceneExplorerTopComponent=The SceneExplorer provides an Overview over the SceneGraph of your Scene.
-SceneExplorerTopComponent.jButton1.text=update
+SceneExplorerTopComponent.jButton1.text=Update
```
capdevon, line 4: Please do not change the button label. Also the label of the **'attach...'** button in the `AppStateExplorer` tab is all lowercase.

capdevon (changes requested):

neph1:
> Thanks for looking into this issue @neph1. I'm curious, does this fix also address the issue of the 'AnimComposer' node and its child animations not updating?

I would think so, but I haven't tested. The original use case worked and I remember specifically adding support in ExternalChangeScanner for the new animation system.

PR #626, opened 2024-11-27
Merge animations actions

neph1:
Based on this discussion: https://hub.jmonkeyengine.org/t/importing-model-and-animations-from-unreal/48015

merge_animations

This includes Wes in the core base libs now. But I don't see it as a bad thing. Might find other uses for it now it's there.

### DIFF
--- a/jme3-core/src/com/jme3/gde/core/assets/actions/MergeAnimationsAction.java
+++ b/jme3-core/src/com/jme3/gde/core/assets/actions/MergeAnimationsAction.java
@@ -42,28 +42,28 @@ import com.jme3.anim.util.HasLocalTransform;
 import com.jme3.gde.core.assets.SpatialAssetDataObject;
 import com.jme3.scene.Node;
 import com.jme3.scene.Spatial;
-import com.jme3.scene.plugins.bvh.SkeletonMapping;
+import com.jme3.util.SafeArrayList;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
-import jme3utilities.MyAnimation;
-import jme3utilities.wes.AnimationEdit;
-import org.bushe.swing.event.Logger;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.openide.DialogDisplayer;
 import org.openide.NotifyDescriptor;
 import org.openide.NotifyDescriptor.Confirmation;
 import org.openide.util.Exceptions;
 
 /**
- * Action for merging one or more spatials' animation to another.
- * Same rig required.
+ * Action for merging one or more spatials' animation to another. Same rig
+ * required.
+ *
  * @author rickard
  */
 public class MergeAnimationsAction implements ActionListener {
-    
+
     private static final String CANCEL = "Cancel";
     private final List<SpatialAssetDataObject> spatials;
     private final Logger logger;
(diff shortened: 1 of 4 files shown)


jme3-core/src/com/jme3/gde/core/assets/actions/MergeAnimationsAction.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+        logger.log(Logger.Level.INFO, "Merging animations done. Saving.");
+        try {
+            targetAsset.saveAsset();
+        } catch (IOException ex) {
+            Exceptions.printStackTrace(ex);
+        }
+    }
+    
+    private static AnimClip retargetClip(
```
capdevon, line 132: The condition `MyAnimation.isJointTrack(animTrack)` is too restrictive and does not consider the case where the target of the `AnimTracks` is a `Node` and not a `Joint` (see animation files downloaded from Mixamo 'without skeleton', which are much more efficient because they take up less disk space). Please modify the code in this way to handle both cases:

 ```java
private static AnimClip retargetClip(AnimClip sourceClip, Armature targetArmature, String clipName) {

	// Create a list to hold the new tracks
	SafeArrayList<AnimTrack> tracks = new SafeArrayList<>(AnimTrack.class);

	// Iterate through each track in the source clip
	for (AnimTrack animTrack : sourceClip.getTracks()) {

		TransformTrack sourceTrack = (TransformTrack) animTrack;
		String targetName = getTargetName(sourceTrack.getTarget());
		Joint target = targetArmature.getJoint(targetName);

		if (target != null) {
			// Clone the source track and set the new target joint
			TransformTrack newTrack = sourceTrack.jmeClone();
			newTrack.setTarget(target);
			tracks.add(newTrack);

		} else {
			logger.log(Level.WARNING, "Joint not found in the target Armature: {0}", targetName);
		}
	}

	// Create a new animation clip with the specified name and set its tracks
	AnimClip newClip = new AnimClip(clipName);
	newClip.setTracks(tracks.getArray());

	logger.log(Level.INFO, "Created new AnimClip {0} with {1} tracks out of {2} from the source clip",
			new Object[]{clipName, tracks.size(), sourceClip.getTracks().length});
	return newClip;
}

private static String getTargetName(HasLocalTransform target) {
	String targetName = null;
	// Determine the name of the target based on its type
	if (target instanceof Node) {
		targetName = ((Node) target).getName();
	} else if (target instanceof Joint) {
		targetName = ((Joint) target).getName();
	}

	return targetName;
}
```

File: jme3-core/src/com/jme3/gde/core/assets/actions/MergeAnimationsAction.java [hunk 2]
```diff
@@ -0,0 +1,208 @@
...
+                new NotifyDescriptor.Confirmation(
+                        "Select spatial to copy animations to.");
+        message.setOptions(spatials);
+        
+        DialogDisplayer.getDefault().notify(message);
+        
+        return message;
+    }
+    
+    private AnimComposer findAnimComposer(Spatial spatial, AnimComposer animComposer) {
```
capdevon, line 179: You can rewrite this method in a much simpler way like this:

Usage: `AnimComposer animComposer = findControl(spatial, AnimComposer.class);`

```java
private static <T extends Control> T findControl(Spatial sp, Class<T> clazz) {
        T control = sp.getControl(clazz);
        if (control != null) {
            return control;
        }
        if (sp instanceof Node) {
            for (Spatial child : ((Node) sp).getChildren()) {
                control = findControl(child, clazz);
                if (control != null) {
                    return control;
                }
            }
        }
        return null;
    }
```

capdevon:
The two modifications I suggested do not use the **Wes** library, but they do the same thing.

jme3-core/src/com/jme3/gde/core/assets/actions/MergeAnimationsAction.java
capdevon: Note: This change optimizes the current solution and is independent of the Wes library.

neph1: Thanks. I've merged these changes and will try them out later. Next time maybe you can create a PR? ;)

jme3-core/src/com/jme3/gde/core/assets/actions/MergeAnimationsAction.java
neph1: I'll save this for later. It might be a good method to have in a central place.

PR #627, opened 2024-11-29
update template libs

neph1:
Added Wes to optional libraries, since it's already in the build.gradle
Found some old version values, although I'm unsure whether they're needed or not. I didn't get the latest Minie when i created a project.

Also, it seems like Heart will be added when selecting Minie as physics lib, but it doesn't appear in the build.gradle. Bug?


tonihele:
It is good to update the old version values just in case. But we do always get the latest automatically (and hope that is compatible with all the stuff around it :D )

PR #629, opened 2024-11-30
Add game template

neph1:
I've had this idea for a long time:
Add some simple game templates for new users to explore. Just a few focused templates. The test project I think is too overwhelming.
This one is actually from that project, a simple marble game about picking up objects, no assets.

More planned: 3rd person game. 1st person game.

What do you think?

image

image

Maybe add a new category  "JME3 - Examples"?


tonihele:
Yeah, great idea. Just that if there is going to be bunch of them, like more than one.. We might want to figure out some method that we can simply extract them from the jME examples so that we don't have to maintain them... Like a hand picked separate catalog of worthwhile jME examples.

neph1:
Maybe we shouldn't bundle them and instead make them downloadable.. through Gradle? Is that possible?

tonihele:
I'm not exactly clear yet on how we should distribute them. But logical starting point would be having these games be separate repos at https://github.com/jMonkeyEngine. Like currently there is the Skull game example. These would be separate from the jme-examples as these would then be more complete examples rather than... almost test case like things.

They would not be coupled with the SDK then also. Which is a good thing.

Maybe to continue from there we would need to create some feature to SDK where we white list these repos and allow user to essentially download the repo as a local copy.

neph1:
MonkeyZone might be a good example to include, too.

neph1:
image

Two more options, Jaimes Ascent and Monkey Zone. Moved all three to a new category - JME3 - Examples.

It's hacky as hell, and not very maintainable atm. I'm sure someone could write a script that does this automatically. Right now it's a lot of duplicated files for a couple of string changes.

tonihele:
Hmm, can it at least be like one code. One wizard, that just takes in the game and whatever as parameters?

neph1:
Yes, it makes sense. I'll see what I can do.

SchmidtServices (approved):

neph1:
I've made an attempt at refactoring this. Maybe I'm doing something wrong, but I can't make it work with a single base template. There's some automagic stuff going on with the netbeans templates it seems, maybe based on package names. It's here if you want a look: https://github.com/neph1/sdk/pull/1/files

tonihele:
I can try to take a look, it would be just so good when thinking about maintainability. Minimal copy paste, just add / remove game from the list and everything is automatic etc

neph1:
What I had in mind in that pr was for each template to extend the base class and supply a config file. I just left the static config file in for testing.

neph1:
@tonihele What's your take on this? It would be nice with a unified structure, but this is the way those modules come out when generated from a "template". Personally, I'm willing to take the bloat it produces. It's contained, after all. As a reminder, there's this pr i started on: https://github.com/neph1/sdk/pull/1
But it doesn't work.

tonihele:
Oh yeah, sorry, I did remember this but I haven't taken a look. I'm also fine it. We can also look later if there is an option to automate it more. But I definitely don't object this PR. It is good to have this feature. So once you decide it is fine, it is good to go.

neph1:
*Spinning up the jMonkeyBrain from cold start*

neph1:
Tested by creating all three projects and running them :rocket: 
Merged main for good measure.
I will merge this during the weekend, unless someone protests.

PR #660, opened 2025-08-24
Feature/issue 611

### LINKED ISSUE #611: Migrate from NBI to nbpackage
As per this message in the [Netbeans Dev Mailing List](https://lists.apache.org/list?dev@netbeans.apache.org:lte=1M:nbpackage), the old Netbeans Installer (NBI) method of packaging Netbeans Platform Applications is essentially deprecated.

We should look to migrate building the jMonkeyEngine SDK installers using the [nbpackage](https://github.com/apache/netbeans-nbpackage/tree/master) utility instead.  

This looks relatively straightforward _in principle_, just need to create the config files and build images with the appropriate tooling (Wine, Swift et al) on which to build and package the .zip file. 

The tricky parts will be:

1. Documentation is pretty sparse ATM
2. The jMonkeyEngine SDK has a custom step/screen in the installer - not sure how to do this yet

This may be a good first step towards converting the SDK to Maven, as the NBI creating in the Netbeans Maven plugin was also not documented super-well (I hadn't figured out how to add the custom screen to the installer...).

The good parts are it should make bundling a JDK easier, as well as potentially solve issues with the installers being trusted/signed on Windows & MacOS

tonihele:
Quite the cleanup. NBI is no more.

Uses NBPackage and heavily leverages Bash scripts. I'm not an expert on either one so there might be some issues there.

Tagging as per usual. Builds ZIP, DEB (as opposed to sh -installer), Windows EXE and MacOS PKG files. MacOS PKGs differ only by bundling the correct JDK. JDKs are automatically downloaded as previously.

After this I suppose it gets a bit easier as we don't need to copy any changed NBI files. This should work regardless of NB version. A bit complicated to understand, hopefully doesn't require so much maintenance...

Current TODO:
- [ ]  Test
- [ ] Figure out how to build PKGs on Linux

Closes #611


tonihele:
Windows installer seems to work as well...

PR #704, opened 2026-06-09
Update to Netbeans 30

tonihele:


tonihele:
I'll merge soon, I don't think anyone will oppose?

PR #706, opened 2026-06-14
Java 25

### LINKED ISSUE #692: Update Java to 25
Update Java to 25. Let's do this once we are preparing for 3.9.0 release. One thing though, those SDK example games... They should be updated as well, otherwise they might not run. SDK then would ship with Java 25 and the example games should use that, also perhaps use updated jME version as well. Of course someone downloading those from older SDK might not get them to work without fiddling with Java versions.

tonihele:
Download and use Java 25 for everything. Resolves #692


tonihele:
Maybe this is everything. Just that the demo projects need to be updated separately to use Java 25 and then the SDK needs to download the tagged releases of Java 25 enabled versions.

@neph1 could you make Java 25 releases of the demo projects? Otherwise they wont run.

neph1:
i've updated the projects (two of them, third is from jme3-tests). i did it blindly as i don't have 25 installed myself. monkeyzone doesn't have releases, and i guess the sdk just gets the latest from master

tonihele:
> i've updated the projects (two of them, third is from jme3-tests). i did it blindly as i don't have 25 installed myself. monkeyzone doesn't have releases, and i guess the sdk just gets the latest from master

Thanks, although they don't work as they include Gradle wrapper that is still 8.6 version. It needs to be updated. i.e., `./gradlew wrapper` once the version is set in the build file.

neph1:
ah crap. i had forgot what the layout looked like. i thought it was the git repos.

tonihele:
> ah crap. i had forgot what the layout looked like. i thought it was the git repos.

No worries, I can also do the ones under jMonkeyEngine since I should have the rights to do so

PR #709, opened 2026-07-18
Revert remove blender

neph1:
Short term fix

The module is even called "GLTF Support" now that I reopen it...


neph1:
I'm having some trouble building the SDK now, which is why this is delayed (and vacation)

